PR #5, opened 2021-08-12
Add "Toggle BOM/CPL" feature

bvernoux:
I think it is convenient in lot of cases to Toggle BOM & CPL at same time in one click so I have added that feature


bvernoux:
I think it is convenient in lot of cases to Toggle BOM & CPL at same time in one click so I have added that feature

Bouni:
Nice! Thanks @bvernoux fro the PR 😄

PR #27, opened 2021-10-08

DanielO:
Avoids needing pandas installed which seems impossible in OSX due to code signing rules.
Uses iconv and sqlite3 binaries to convert the CSV rather than using Python for performance.

### DIFF
--- a/library.py
+++ b/library.py
@@ -1,6 +1,9 @@
+import csv
+import datetime
 import locale
 import logging
 import os
+import os.path
 import time
 from pathlib import Path
 import requests
@@ -13,8 +16,8 @@ class JLCPCBLibrary:
     def __init__(self, parent):
         self.parent = parent
         self.create_folders()
-        self.csv = os.path.join(self.xlsdir, "jlcpcb_parts.csv")
-        self.db = os.path.join(self.xlsdir, "jlcpcb_parts.db")
+        self.csvfn = os.path.join(self.xlsdir, "jlcpcb_parts.csv")
+        self.dbfn = os.path.join(self.xlsdir, "jlcpcb_parts.db")
         self.loaded = False
         self.logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


Bouni:
Ok, I did some test on Windows. I put the necessary binarys and dlls in a bin folder within the plugin folder an got it working.
But it needs quite some changes:

These are the lines I've modified: https://github.com/Bouni/kicad-jlcpcb-tools/blob/39be32c7a8d29418fda9b65dac305b5ef080ce71/library.py#L64-L65

```python
        iconv_path = os.path.join(self.bindir, "iconv.exe")
        with open(utf8_path, "w") as f:
            subprocess.call(
                [iconv_path, "-f", "gbk", "-t", "UTF-8", self.csv], stdout=f
            )
        os.chdir(self.bindir)
        subprocess.call(
            [
                "sqlite3.exe",
                "../jlcpcb/jlcpcb_parts.db",
                "-cmd",
                f".import --csv '../jlcpcb/jlcpcb_parts_utf8.csv' jlcpcb_parts",
            ],
            stdin=subprocess.DEVNULL,
            stderr=subprocess.DEVNULL,
        )
```
Maybe we can somehow pipe these with subproccess popen / communicate magic 🧙🏽 

These are the .exe files and the dll files, mostly from the GnuWin repo on sourceforge:

```
iconv.exe
libcharset1.dll
libiconv2.dll
libintl3.dll
sqlite3.exe
```

I've not yet looked into the licences to judge if we can simply ship them with the plugin ...

For sure we need to build some kind of os switch logic into this in order to get it working.

DanielO:
> Ok, I did some test on Windows. I put the necessary binarys and dlls in a bin folder within the plugin folder an got it working. But it needs quite some changes:
> 
> These are the lines I've modified:
> 
> https://github.com/Bouni/kicad-jlcpcb-tools/blob/39be32c7a8d29418fda9b65dac305b5ef080ce71/library.py#L64-L65
> 
> ```python
>         iconv_path = os.path.join(self.bindir, "iconv.exe")
>         with open(utf8_path, "w") as f:
>             subprocess.call(
>                 [iconv_path, "-f", "gbk", "-t", "UTF-8", self.csv], stdout=f
>             )
>         os.chdir(self.bindir)
>         subprocess.call(
>             [
>                 "sqlite3.exe",
>                 "../jlcpcb/jlcpcb_parts.db",
>                 "-cmd",
>                 f".import --csv '../jlcpcb/jlcpcb_parts_utf8.csv' jlcpcb_parts",
>             ],
>             stdin=subprocess.DEVNULL,
>             stderr=subprocess.DEVNULL,
>         )
> ```

This quick test worked for me in Windows:
```python
sqlite = 'y:\\downloads\\sqlite-tools-win32-x86-3360000\\sqlite3.exe'
iconv = 'y:/downloads/gettext0/bin/iconv'
csv = 'y:/downloads/jlcpcb_parts.csv'
subprocess.check_call([sqlite, '-cmd', '.import --csv "|%s -f gbk -t UTF-8 %s" jlcpcb_parts' % (iconv, csv), 'c:\\temp\\test.db'], stdin = subprocess.DEVNULL, stderr = subprocess.DEVNULL)
```
Note the iconv and csv path need to be forward slashes otherwise they get munched going through sqlite, I tried escaping the back slash version (passing them through shlex.quote) but it failed with a different error:

```python
sqlite = 'y:\\downloads\\sqlite-tools-win32-x86-3360000\\sqlite3.exe'
iconv = 'y:\\downloads\\gettext0\\bin\\iconv'
csv = 'y:\\downloads\\jlcpcb_parts.csv'
subprocess.check_call([sqlite, '-cmd', '.import --csv "|%s -f gbk -t UTF-8 %s" jlcpcb_parts' % (shlex.quote(iconv), shlex.quote(csv)), db], stdin = subprocess.DEVNULL)
The filename, directory name, or volume label syntax is incorrect.
<pipe>: empty file
```
> Maybe we can somehow pipe these with subproccess popen / communicate magic 🧙🏽

I had a look but I can't see how to get sqlite's .import command to read from stdin (on windows anyway, on Linux/OSX /dev/stdin would probably work)

IMO it's not a deal breaker if we have to run iconv and sqlite serially - it is a bit slower but not that much.

> I've not yet looked into the licences to judge if we can simply ship them with the plugin ...

Sqlite is public domain: https://www.sqlite.org/copyright.html
The iconv libraries are LGPL but the binary is GPL: https://www.gnu.org/software/libiconv/ (look for "copyright")
There is win-iconv (https://github.com/win-iconv/win-iconv) which uses Win32 APIs and is public domain but I have not tested it.

> For sure we need to build some kind of os switch logic into this in order to get it working.

Yeah, I think detecting Windows is pretty easy though.

PR #28, opened 2021-10-09

DanielO:
This is a reworked version of my previous attempt and is pure python so should work on any platform.


DanielO:
This has a bug..

PR #29, opened 2021-10-09

DanielO:
This is a pure python implementation (and fixed the bug where I forgot to flush the remaining rows at the end of the CSV)

### DIFF
--- a/library.py
+++ b/library.py
@@ -1,34 +1,23 @@
 import csv
+import datetime
 import locale
 import logging
 import os
+import os.path
 import time
 from pathlib import Path
-
-# Hack to avoid wx + pandas error "ValueError: unknown locale: en-GB"
-try:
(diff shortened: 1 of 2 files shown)


Bouni:
Ideally we get rid of the `wx.BusyInfo` and use the progress bar instead (https://github.com/Bouni/kicad-jlcpcb-tools/blob/main/plugin.py#L234 for example).

That would make the UI much more consistent in my opinion!

DanielO:
> > That is my "all in 1" branch, it has some other changes I'd like to push too.
> I would prefer a PR for each change unless they're just minor cleanups, that way its clear what changes are made.

Sure thing, I just meant you could have a sneak preview😂

I can cherrypick stuff into separate branches although I can't see how to make them serial in GitHub as yet so I'll do them one at a time.

PR #31, opened 2021-10-11
Show datasheet URL and allow user to open it in a browser.

DanielO:
Like it says on the tin.


DanielO:
Huh, nope... For me it's normal sized - I haven't much wx experience though. I'll see if I can fix it.

DanielO:
I think I might have fixed it but you'll have to check.

DanielO:
I've updated it again, this time for sure (maybe).

DanielO:
> Have you put that line into your PR? I don't see it in the changes!?

Sorry, I went to bed. Your suggestion works on OSX so I've updated the PR.

PR #32, opened 2021-10-11
Split keywords like shell (with quoting) and match each one.

DanielO:
All keywords must be present (in any field) to match.
This makes it easy to search for (say) 10uF 25V capacitors because the order those words are in varies by part.


Bouni:
This is a really nice one! 😎

PR #33, opened 2021-10-11
Add keyboard shortcuts for closing windows and search on enter

DanielO:
Also sets focus on the keyword text box when the window is created.


DanielO:
Good idea, done.

Bouni:
Thanks for all you brilliant PRs!

PR #37, opened 2021-10-11
Download and convert the CSV directly in another thread.

DanielO:
- Fixes UI being unresponsive.
- Disable buttons during download.
- Display the filename and part count when loaded.
- Fix loading at start.
- Remove wx.BusyInfo since it doesn't seem to do anything.

### DIFF
--- a/library.py
+++ b/library.py
@@ -1,15 +1,16 @@
 import csv
-import datetime
 import locale
 import logging
 import os
 import os.path
 import time
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Bouni:
On windows I see some minor UI hickups, I experianced such things if wx elements are not assigned to a panel as far as I remember:

grafik

DanielO:
OK that works for me, I've updated the PR (27e6019).

Issue #42, opened 2021-10-21
PermissionError on attempt to delete database


Bouni:
At least on Windows I get an `PermissionError` when the database file should get deleted.

PR #43, opened 2021-10-21
Utilize contextmanger to prevent non closed db handles

Bouni:
As described in #42 in some situations the db file cannot be deleted because the file is blocked by the other thread.
In order to prevent this, I refactored the library part to utilize contextmanger.

@DanielO would you mind taking a look at what I did and test it?


Bouni:
@DanielO Thanks for your help, highly appreciate it 😄

PR #67, opened 2021-10-29
Export unLCSCed components in BOM

### LINKED ISSUE #65: Add option to include non-selected components in BOM
Currently, all components have to be selected in order to be included in BOM. However, most resistors and capacitors can be automatically matched in JLC, making it unnecessary (and painful) to select them manually. (The slow, laggy UI adds extra pain in the ass if there are 30's individual values of resistors: every select makes the UI freeze for 3 seconds)

t123yh:
Close #65


Bouni:
I'll close this in favor of #69

PR #70, opened 2021-11-03
[WIP] Calculate correct part prices using the JLC API

Bouni:


Bouni:
I'll close tis and lokk into it somewhere in the future

PR #71, opened 2021-11-04
improve startup speed by using thread to get stock info

Bouni:
Using [pyinstrument](https://pyinstrument.readthedocs.io/en/latest/home.html) I was able to see that querying the stock info from the database for each part took a lot of time and slowed the startup massively.

I'v put that into a seperate thread which mand the startup nice and fast again 🥳

### DIFF
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import os
 import sys
+from threading import Thread
 
 import wx
 import wx.dataview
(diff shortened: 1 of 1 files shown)


Bouni:
This is not ideal as the stock is refreshed every time we toggle a BOM value for example.
I think a bigger overhaul is neccessary to get this solved properly.
I'm going to close this PR and submit a fresh one in the coming days!

PR #79, opened 2021-12-04
Added zone refill with user prompt

simply-space:
This PR prompts the user if they wish to refill zones when they generate assembly files (This is the behavior I got used to with the built in plot gerber tool and I got burnt badly when I switched to this plugin).

I couldn't find a way to detect if an update of the zones is needed, so this will always prompt the user. 

Using filler.Fill(zones, True) does generate a prompt if it detects the zones aren't up to date, however the cancel button on that doesn't appear to stop the zones filling. I'm not sure how to implement this, or if it is just a bug in the PCBnew API.

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -11,6 +11,8 @@ from pcbnew import (
     PCB_PLOT_PARAMS,
     PLOT_CONTROLLER,
     PLOT_FORMAT_GERBER,
+    ZONE_FILLER,
+    Refresh,
     B_Cu,
     B_Mask,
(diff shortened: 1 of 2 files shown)


Bouni:
Strange, if I call the fill zones with `aCheck` set to True, I'll get an automatic prompt

```python
def fill_zones(self):
    """Refill copper zones following user prompt."""
    filler = ZONE_FILLER(self.board)
    zones = self.board.Zones()
    filler.Fill(zones, True)
    Refresh()
```

image

Issue #87, opened 2021-12-16
Exception when trying to use the tool


nickoe:
I get the following exception, not sure what to make of it. A couple of weeks ago I was able to start the script. I have tried to uninstall and install again.
![image](https://user-images.githubusercontent.com/352202/146395016-54df9bc3-54e0-4635-a607-fdd7c002cf45.png)

I also recently updated my system, including a ["rebuild"](https://github.com/archlinux/svntogit-community/commits/packages/python-wxpython/trunk) of wxpython
```
[2021-12-15T20:49:56+0100] [ALPM] upgraded python-wxpython (1:4.0.7.2-1 -> 1:4.0.7.2-3)
```

Kicad version info.
```
Application: KiCad PCB Editor

Version: (6.0.0-rc2), release build

Libraries:
	wxWidgets 3.0.5
	libcurl/7.80.0 OpenSSL/1.1.1l zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 libidn2/2.3.2 libpsl/0.21.1 (+libidn2/2.3.0) libssh2/1.10.0 nghttp2/1.46.0

Platform: Linux 5.15.6-arch2-1 x86_64, 64 bit, Little endian, wxGTK, i3, x11

Build Info:
	Date: Dec 16 2021 15:43:50
	wxWidgets: 3.0.5 (wchar_t,wx containers,compatible with 2.8) GTK+ 3.24
	Boost: 1.76.0
	OCC: 7.5.3
	Curl: 7.80.0
	ngspice: 35
	Compiler: Clang 13.0.0 with C++ ABI 1002

Build settings:
	KICAD_USE_OCC=ON
	KICAD_SPICE=ON
```

Bouni:
I think I remember having the same (or at least similar) problem on my Arch notebook when ever I've upgraded my system which upgraded wxpython to a version greater than 4.0.7.something 🤔 

Let me double check that. Maybe you can try downgrading that and giving it a try.

Bouni:
I can confirm that I see the same problem on my Arch notebook but downgrading wx didn't help.
I'll try to figure out how to get this working again.

Bouni:
I found a potential solution for this problem, but I'll wait for @DanielO to test this on OSX to prevent creating a regression there before merging this into main.

Bouni:
I'm going to merge this as its a show stopper for linux users. If this is a regression on OSX I'll have look into it when somebody creates an issue about it.

PR #88, opened 2021-12-17
Potential fix for #87

Bouni:
There is a problem with the AcceleratorTable in recent KiCad versions on Linux.
With these changes it works on Linux as well as on Windows 10.
OSX still needs testing, @DanielO would you mind helping me aut with a OSX test, please?


DanielO:
Hi, sorry for the delay!
Unfortunately it crashes with what looks like a GUI call from a non-main thread.
I will file a bug about it.

PR #99, opened 2022-01-09
Write the POS file as utf8

obra:
Before this change, parts that have a description containing non-ascii symbols, such as a +- or a µ would cause writing of the .pos file to silently fail.


Bouni:
Thanks for your PR!

PR #112, opened 2022-01-22
Show main window as modeless dialog instead of modal

eric-wood:
When verifying my BOM before generating the various fabrication files I often find myself wanting to 
view other KiCad windows, like eeschema, which I cannot do while the JLC tools modal is open. With this change it's possible to cycle through other KiCad windows while this plugin's dialog is open.

This may be a personal preference thing, so feel free to disregard; it's an easy change to make and I don't mind patching my own version, but I figured other people would find this helpful as well!

Thanks again for this wonderful plugin, it has saved me so much time!

### DIFF
--- a/plugin.py
+++ b/plugin.py
@@ -19,5 +19,4 @@ class JLCPCBPlugin(ActionPlugin):
     def Run(self):
         dialog = JLCBCBTools(None)
         dialog.Center()
-        dialog.ShowModal()
-        dialog.Destroy()
+        dialog.Show()


Bouni:
I thought about doing this a couple of times but always hesitated because I'm not sure if things work as expected if the user gets any chance to modify the board in the background.
But on the other hand I wanted to lookup thing the way you described it and was blocked by the modal.
I'm going to merge this PR and we'll see if and how many people get in trouble by making changes while the plugin is open.

Bouni:
I don't know why but github merged this and still shows me a merge conflict :raised_eyebrow:

Bouni:
I consider this the schrodinger's pull request and close it 😆

PR #133, opened 2022-02-08
Handle part select function when multiple rows are selected (no crash anymore)

bhusang:
I have seen a nasty crash of Kicad when multiple rows are selected and then the "Select Alike" button is pressed. Now the code just reports that you need to select only one row.


Bouni:
Lokks good, thanks for your PR!

PR #137, opened 2022-02-11
added Copy LCSC to and Paste LCSC from the clipboard buttons

computergeek1507:


Bouni:
Nice feature, thanks for the PR 😁!

I'll merge it but with this I realized I need to come up with some sort of new toolbar, maybe a right click menu or something like that:

grafik

Any Idea is highly appreciated! (Create an issue if you have suggestions)

PR #138, opened 2022-02-11
Added rotation offset column to component list

computergeek1507:
Clear "Regex" value on list deselection to re-enable "Add" functionality


Bouni:
Nice one, thanks!

I really want to have multiline captions so that this could be Rotation Offset but it's not possible to my knowledge 😞 

grafik

Maybe I find a solution ... wx lacks features in so many places

Issue #160, opened 2022-04-17
Enable to add custom layers (v-score for example)


AntonKrug:
JLPCB allows custom panelization which can be done on the user's side. They allow users to provide custom layer, for example vscore and they grove it out when given instructions to do so. However there is no way in this plugin to let consider (generate such layers)

**Describe the solution you'd like**
Have a 'manage layer' button which would could have checkboxes which layers would be considered. KiKit panelization plugin by default outputs of the v-cut information into the 'User.Comments' layer, this could be the default layer to be considered.

Bouni:
Is a custom selection actually needed or would it be sufficiant to just output the `User.Comments` by default?
A additional exported layer would not hurt even if its empty, right?

Bouni:
Resolved by #161

PR #161, opened 2022-04-18
Export User.Comments layer as a V-score layer

AntonKrug:
The KiKit panelize plugin export the V cut marking into the User.Comments layer by default. JLCPCB support custom V scoring specified by the end-user just it needs to be exported in own gerber file and mentioned in the job comment.


Bouni:
😄 Just wrote a proposal in your issue #160 and then realized that you already sent exactly that as a PR ...

Issue #167, opened 2022-05-16
Untented vias not supported/honored


Engineer8468:
In my "Fabrication Outputs" dialog I have "Do not tent vias" checked, but the gerber outputs still tent the vias.  Can this plugin be changed to respect this setting?

Bouni:
Sure, can you test if the the [PR #168](#168 ) solves your problem?

Bouni:
@livigni Sorry for the delay. I think I need to add some sort of config option to allow users to select if they want tented vias or not.
AS a quick workaround you can edit this line https://github.com/Bouni/kicad-jlcpcb-tools/blob/main/fabrication.py#L114 from `True` to `False`

ColinBathe:
Can I add a vote to make this optional please!

It would also be useful to know where the fabrication.py file ends up on windows? I can't find it!

Thanks. Love the plugin.

Bouni:
Just to keep you up to date, this is WIP!
I started changing all the buttons for toolbars as the make my life much easier. However that's a lot of work and takes some time.

Here's a quck peek:

![image](https://user-images.githubusercontent.com/948965/181559301-ac6b0008-3b10-4409-81f5-a00f51d06197.png)

And the settings menu with experimental toggle switches:

![image](https://user-images.githubusercontent.com/948965/181560363-8d1675c8-0ca9-4038-91cc-bcae3758566b.png)

Not sure if I go further down that path or simply use checkboxes 🤷🏽

Oliver0804:
Hi sir,
If I change this line to false, can I change the via to have the ink cover the holes?
      
 popt.SetPlotViaOnMaskLayer(True) # Do not tent vias

I look forward to you adding this feature to the export options..<3

Bouni:
> Hi sir,
> If I change this line to false, can I change the via to have the ink cover the holes?
>       
>  popt.SetPlotViaOnMaskLayer(True) # Do not tent vias
> 
> I look forward to you adding this feature to the export options..<3
> 

Yes, exactly!

riskable:
Just want to confirm that my gerber files generated via JLCPCB tools are generating without tented vias which doesn't match my plotting settings.  It seems that no matter what my settings are the tools still get generated with "Do not tent vias" enabled (the opposite of what I want).

Bouni:
[...]

PR #168, opened 2022-05-17
Add 'do not tent vias' option to fabrication

Bouni:
This adds the "Do not tent vias" option to the generate fabrication files function as requested in #167


Engineer8468:
Thanks! I will try to test this.  I admit I'm a bit of an old-schooler and a hardware guy, so github is unfamiliar.

PR #174, opened 2022-06-08
Added Footprint/Value to LCSC Partnumber "Mapping" database

computergeek1507:
Created a Footprint/Value to LCSC Partnumber "Mapping" database and dialog. Mapping can be saved and reused with new projects.

### DIFF
--- a/library.py
+++ b/library.py
@@ -224,6 +224,61 @@ class Library:
                     ).fetchall()
                 ]
 
+
+    def create_mapping_table(self):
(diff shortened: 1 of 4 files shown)


Bouni:
Hey Scott, brilliant as always, just gave it a try, works like charm 😎 
I'll merge this and add icons later today if I find some time.

computergeek1507:
I thought about running the mapping lookup on startup, but I don't know if I that was a good idea

PR #183, opened 2022-06-24
added exporting LCSC property to KiCad v6 Schematic Files

computergeek1507:
This is very experimental, I wanted to get feedback if it is a maintainable feature. I tested it on a couple of my designs so far.

### DIFF
--- a/helpers.py
+++ b/helpers.py
@@ -13,6 +13,12 @@ EXCLUDE_FROM_BOM = 3
 NOT_IN_SCHEMATIC = 4
 
 
+def getWxWidgetsVersion():
+    v = re.search(r"wxWidgets\s([\d\.]+)", wx.version())
+    v = int(v.group(1).replace(".", ""))
+    return v
+
+
 def getVersion():
     """READ Version from file"""
     if not os.path.isfile(os.path.join(PLUGIN_PATH, "VERSION")):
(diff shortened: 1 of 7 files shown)


Bouni:
Wow, that looks very good 😍 
Give me a bit time to test this before I merge it!

Bouni:
@computergeek1507 I did a quick test and it seems to work as expected! 

But it fooled me on te very first try 🙈 I clicked the export button selected the .kicad_sch file and clicked ok.
So far so good I thought and wanted to see in the schematic if the LCSC property is set, but it wasn't.
Then I realized that the FileDialog gave me another project as default directory and I wasn't reading properly when I selected that  .kicad_sch file. That way the function wrote data from one projects pcb into the schematic of another one.
Sure, there's the _old file as backup but still I think this should be tune a bit.

In my opinion at least the defaults of the FileDialog should be set to reasonable values:

```python
with wx.FileDialog(
            self,
            "Select Schematics",
            "", # DefaultDir
            "", # DefaultFile
            "KiCad V6 Schematics (*.kicad_sch)|*.kicad_sch",
            wx.FD_OPEN | wx.FD_FILE_MUST_EXIST | wx.FD_MULTIPLE,
            ) as openFileDialog:
```

- DefaultDir should be the project dir
- DefaultFile could be the same name as the .kicad_pcb but with kicad_sch

Let me know what you think about this.

Bouni:
@computergeek1507 Holy 💩 I tried to add my suggestions but messed it up 🙈 I think I simply push a new branch with your changes and add mine, otherwise I'm going to mess up the main branch.

Issue #194, opened 2022-07-07
Component offset in POS file


tmckzs:
Hi,
To be fair, I am not sure if it is some configuration I have, or something to do with the plugin..

**Describe the bug**
Position data is offset for components selected through the JLCPCB component selector. ~7.6mm in -x direction

**To Reproduce**
1. Go to PCBNEW
2. Click plugin button
3. Select JLCPCB Components
4. Generate assembly files
5. View POS file is incorrect (Within JLCPCB viewer) -> components are shifted 7.61118mm -x

**Expected behavior**
POS file to be correct

# KiCad Version
```
Application: KiCad (64-bit)

Version: (6.0.6), release build

Libraries:
	wxWidgets 3.1.6
	libcurl/7.82.0-DEV Schannel zlib/1.2.12

Platform: Windows 11 (build 22000), 64-bit edition, 64 bit, Little endian, wxMSW

Build Info:
	Date: Jun 26 2022 02:39:06
	wxWidgets: 3.1.6 (wchar_t,wx containers)
	Boost: 1.79.0
	OCC: 7.6.0
	Curl: 7.82.0-DEV
	ngspice: 37
	Compiler: Visual C++ 1929 without C++ ABI

Build settings:
	KICAD_USE_OCC=ON
	KICAD_SPICE=ON

```

Bouni:
7.6mm sounds dangerously close to 7.62mm which is 300mil, might be a inch/metric setting in KiCAD 🤔 
Can you share the project that makes the problem?

tmckzs:
Yes, I will send it to you.

v/r
Tharon Stewart
719-964-8534

tmckzs:
Interesting, for me it offsets R1..R4 -x ~7.62mm.

Version is 2022.01.01

I suppose one of my local settings must be conflicting... thank you for your investigation - will try resetting KiCad settings, etc.

Jan21493:
I have a similar problem, maybe it's related to the "origin point" that I had set to the upper left corner of the pcb (recommendation from a youtube video to find the position on the board easier). The gerber file looks to be O.K., also the viewer at JLCPCB had correct placements. Only the assembly of parts has reported an error:

![Produce_DanZhi SMT_Original_Snapshot Top 3578953A_Y2 SMT0220726917126](https://user-images.githubusercontent.com/25911411/181204383-163454da-2229-4cb0-8b6d-3bcedc8d986b.png)

[...]

PR #199, opened 2022-07-28
Take AuxOrigin into accout when generating POS file

Bouni:
As suggested by @Jan21493 in #194 taking the Aux Oriin into account when generating the POS file should fix position offsets for those who changed their KiCAD Grid origin settings 🚀

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -231,7 +231,7 @@ class Fabrication:
         """Generate placement file (POS)."""
         posname = f"POS-{self.filename.split('.')[0]}.csv"
         self.corrections = self.parent.library.get_all_correction_data()
-
+        aux_orgin = self.board.GetDesignSettings().GetAuxOrigin()
         with open(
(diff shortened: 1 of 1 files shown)


Jan21493 (approved):
That was quick! Yes, it works as expected the positions are set to the aux origin point!

However I still have the problem that the positions were wrong for smd assembly. I'll add everything to #194.

Jan21493 (changes requested):
I did some further testing and have shared the results in #194 : So far it does not work as expected, because the gerber files still use the upper left corner of the page as the origin point. 

I had another look into the **_fabrication.py_** file and the "inspired by ..." link. There I found the missing command: `popt.SetUseAuxOrigin(True)` That needs to be added to the "# gerber section", e.g. at the end at line 126. 

With that command added, the gerber files and POS aka Component Placement List (CPL) file are using the same origin and it works fine.

PR #202, opened 2022-08-03
minor bugfix: don't overwrite duplicate column name

binary1230:
- line 247 incorrectly overwrites self.lcsc with a new column named 'Type', erasing it. line 240 is in the right here
- to fix, we just store this in a different variable
- this wasn't affecting anything because it wasn't used by anything but, was confusing and maybe someday could cause some future developer some heartache


binary1230:
hm.... sorry it appears I have some kind of line ending issue. this is a 1-line code fix, but it's flagged the entire file as changed. woops. will resubmit.

binary1230:
I will re-create this

PR #204, opened 2022-08-03
convert mainwindow.py from CRLF line endings to LF

binary1230:
this is legit
- this one python script (mainwindow.py), is for some reason checked into the github repo with CRLF line endings in it, causing merging/pull requests made it against it to be weird and cause a bunch of chaos. all the other python scripts are already LF.
- merging this pull request will do nothing except convert CRLF to LF, which is what you almost certainly want unless there's a specific reason for it [even under windows deploys]

I have two other pull requests that will be applied on top of this commit. I suggest if you want these patches, apply THIS pull request first, then apply everything else. it'll make reviewing the diffs easier.

if not, another option is to hit the 'toggle show all whitespace' button in github which will filter out CRLF so you can read the diffs.

thanks for the awesome project btw.


Bouni:
This CRLF / LF fuckup is most likely du to the fact that I developed on Windows and Linux and messed that up somehow 🤷🏽‍♂️

PR #205, opened 2022-08-03
Fix parts modal dialog opening issues

binary1230:
please merge https://github.com/Bouni/kicad-jlcpcb-tools/pull/204 first, this PR is built off that.  if you don't, the line ending conversion will make the diff look weird (you can toggle 'show whitespace' in github to work around it though)

bugfix: if you clicked 'Show part details' on a part that didn't have a valid LCSC # entered, a couple of weird things would happen. this PR fixes and smooths over some of that.

Details:

1. fix unending busy cursor: this wasn't hurting anything but feels very broken when it happens: if an exception is raised in the 'View Parts Detail' modal dialog (like, when it can't find an LCSC part#), then the original code would never execute 'del self.busy_cursor' to stop the busy cursor. users would be able to use the app normally but, it would never unset the busy cursor.  fix by using a try/finally to call wx.BeginBusyCursor() and always esnure that wx.EndBusyCursor() is called.
- further work could be done to extract this to a generic "dialog opening" function
- I think I saw an old bug or forum post about this a while ago

2. avoid looking up column names by index, instead look them up by title. that way, the lookup is future-proof and not error-prone (i.e. coders shouldn't have to know that "LCSC" is column "3" [today, in 2022], they can now call a series of helper functions that iterate through the columns and look it up. if performance is ever an issue (there's no way, but), this could be cached at app startup. this change provides the scaffolding needed to retrofit the app so harcoding column IDs isn't necessary anymore.
- further work could cache these at registration time so the column title is only ever a string in one place in the code and used for everything

screenshot of the (at least a little) bit nicer looking UI giving some hints about how to correct:
image


mainwindow.py
```diff
@@ -1,1057 +1,1093 @@
...
+
+        return lcsc_ids_selected
+
+    def get_row_item_in_column(self, row, column_title):
+        return self.footprint_list.GetTextValue(row, self.get_column_position_by_name(column_title))
+
+    def show_part_details_dialog(self, part):
+        wx.BeginBusyCursor()
+        try:
+            self.logger.info(f"Opening PartDetailsDialog window for part with value: '{part} (this should be "
```
Bouni, line 895: Is this a dev artifact or is there a reason why that should be logged whenever the part details dialog is opened?

binary1230: Probably a bit of a dev artifact at this point. I think I started this trying to just log what was causing the issue, then ended up doing the rest of this. It could be commented out or removed now, sure.  Was vaguely thinking if there were problems in the future or with http retrieval, this could provide a breadcrumb for users to either fix a busted part# or to file a bug report where the log would contain more useful info right before stuff went wrong 

I'm good either way, if you think it's too spammy let's remove it

Bouni: I think I'll comment it out

PR #208, opened 2022-08-10
Fix schematic export

fredszaq:
The condition was inverted and prevented from finding the part in the store. Inverting it makes the export work


Bouni:
Thanks for your PR!

PR #213, opened 2022-08-25
schematic export fixes

computergeek1507:
skip symbol library part section, fix single digit XY positions, fix zero is a valid property ID, fix add/replace logic to prevent duplicate LCSC fields, added more logging, delete temp "old" files


Bouni:
Thanks for the fix!

PR #214, opened 2022-08-28
Added info about how to run pluggin to READEME.md

jevonlongdell:
This could be very useful for people new to KiCAD. It would have saved me a few minutes :)


Bouni:
Thanks for your PR!

PR #216, opened 2022-09-02
Change the Category hint to match the parts database

ajf58:
Using the example 'Resistor' gives no results. Use the plural form 'Resistors' instead.


Bouni:
@ajf58 Thanks for the PR!

PR #218, opened 2022-09-07
Offset drill file to Aux origin to match other fabrication files

msalau:
When Aux origin is set, generated drill files don't match gerbers and position files. This inconsistency happens due to offset calculation. Gerbers and position files use Aux origin for offset, while drill files use fixed (0, 0) for all cases.

This can be easily seen on the screenshot (aux origin is set to lower left corner of the PCB):

image

The pull-request makes offset for drill files identical to gerbers and position files.


Bouni:
Thanks for your PR!

PR #227, opened 2022-09-23
Support selecting category and subcategory from a ComboBox

ajf58:
This PR adds the ability to select the category and, optionally, subcategory from a dropdown list populated from the parts database. An example screenshot is shown below.

image


Bouni:
This is a nice one 😍 Thank you for this PR!

PR #235, opened 2022-11-01
escape footprint names when using them to make a regex

venkatrao1:
Footprints with names like Connector(2-pin) will generate regexes that don't actually match the footprint name when you click "Add Rotation". This fixes that by escaping the special characters, at least for the connector in my layout.


Bouni:
Thank you for the PR

Issue #241, opened 2022-11-26
Not Able to Download JLC Library Database (404)


abish7643:
**Describe the bug**
I tried to update to the latest version of kicad-jlcpcb-tools, upon using it, there's an error as shown in the image.

**To Reproduce**
1. Go to PCB Editor
2. Click on the Plugin button
3. Click on Update Library

**Expected behavior**
Automatically downloading LCSC Part Library.

# KiCad Version

```
Application: KiCad

Version: 6.0.9-8da3e8f707~117~ubuntu22.10.1, release build

Libraries:
	wxWidgets 3.0.5
	libcurl/7.85.0 OpenSSL/3.0.5 zlib/1.2.11 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.3 libpsl/0.21.0 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.49.0 librtmp/2.3

Platform: Linux 5.19.0-23-generic x86_64, 64 bit, Little endian, wxGTK, ubuntu, x11

Build Info:
	Date: Nov 24 2022 00:22:25
	wxWidgets: 3.0.5 (wchar_t,wx containers,compatible with 2.8) GTK+ 3.24
	Boost: 1.74.0
	OCC: 7.6.3
	Curl: 7.85.0
	ngspice: 37
	Compiler: GCC 12.2.0 with C++ ABI 1017

Build settings:
	KICAD_USE_OCC=ON
	KICAD_SPICE=ON
```

abish7643:
![JLCPCB_KicadPlugin_Issue](https://user-images.githubusercontent.com/42410225/204075150-87869a18-c0cb-4df8-b7ec-aa886c04f285.png)

Bouni:
404 means that JLC might changed the download link to their database.
I'll take a look on Monday

abish7643:
Makes sense. Thanks for the quick response.

binary1230:
yea, def the root cause of #238 , I looked around and couldn't find an equivalent CSV download, though, I don't really know where to look.

in a pinch, I would love to use the rest of the plugin even without the stock levels or parts data, just because the POS/BOM export/etc is still really useful.  

I'm having a look to see if we can catch some of the sql exceptions and continue on even if the URL 404's out (unless you think that's just a silly thing to attempt)

binary1230:
[...]

Issue #238, opened 2022-11-08
Tool not working in KiCAD 6.0.9 on MacOS


BenViete:
**Describe the bug**
After installing the plugin (plugin manager / GitHub clone) I try to start the plugin. If there is no jlcpcb folder in my project folder it starts, but has no parts listet on the popup window. If I close the window and try to start the plugin again I get the following error message:

```
Traceback (most recent call last):

File "/Users/vieteben/Documents/KiCad/6.0/scripting/plugins/kicad-jlcpcb-tools/plugin.py", line 21, in Run
dialog = JLCPCBTools(None)

File "/Users/vieteben/Documents/KiCad/6.0/scripting/plugins/kicad-jlcpcb-tools/mainwindow.py", line 571, in __init__
self.init_store()

File "/Users/vieteben/Documents/KiCad/6.0/scripting/plugins/kicad-jlcpcb-tools/mainwindow.py", line 587, in init_store
self.populate_footprint_list()

File "/Users/vieteben/Documents/KiCad/6.0/scripting/plugins/kicad-jlcpcb-tools/mainwindow.py", line 660, in populate_footprint_list
details = self.library.get_part_details(numbers)

File "/Users/vieteben/Documents/KiCad/6.0/scripting/plugins/kicad-jlcpcb-tools/library.py", line 318, in get_part_details
return cur.execute(

sqlite3.OperationalError: no such table: parts
```

**To Reproduce**
1. Install plugin in KiCAD 6.0.9 on MacOS
2. Click on JLCPCBTools button
4. See error message (exception)

**Expected behavior**
Tool should run as expected.

# KiCad Version
```
Application: KiCad Leiterplatteneditor
Version: (6.0.9-0), release build
Libraries:
    wxWidgets 3.1.5
    libcurl/7.79.1 SecureTransport (LibreSSL/3.3.6) zlib/1.2.11 nghttp2/1.45.1
Platform: macOS Version 12.6.1 (Build 21G217), 64 bit, Little endian, wxMac
Build Info:
    Date: Oct 29 2022 19:31:49
    wxWidgets: 3.1.5 (wchar_t,wx containers)
    Boost: 1.80.0
    OCC: 7.6.3
    Curl: 7.64.1
    ngspice: 37
    Compiler: Clang 12.0.0 with C++ ABI 1002
Build settings:
    KICAD_USE_OCC=ON
    KICAD_SPICE=ON
```

Bouni:
[...]

PR #242, opened 2022-11-26
[emergency fix] attempt to allow startup even if we can't download the database

binary1230:
JLCPCB's CSV URL stopped working 5 days ago, and the plugin hasn't really had to handle this case so things can get weird. This fix attempts to work around and re-enable exporting and other features while we figure out a more long-term fix (better data source etc).

See #241 and #238 for more info

This is far from a comprehensive and well-tested "offline mode", but it gets the job done for me enough to export fab files, BOM, and POS files.  This could probably use a lot of testing.

future work:
- this could use some more UI polish to indicate better what's going on and what the consequences of that will be. for now it works as an emergency fix.
- pull using a cached data source (I have one, will try and work that into a future PR)

### DIFF
--- a/library.py
+++ b/library.py
@@ -219,12 +219,11 @@ class Library:
         """get all corrections from the database."""
         with contextlib.closing(sqlite3.connect(self.dbfile)) as con:
             with con as cur:
-                return [
-                    list(c)
-                    for c in cur.execute(
-                        f"SELECT * FROM rotation ORDER BY regex ASC"
-                    ).fetchall()
-                ]
+                try:
+                    result = cur.execute(f"SELECT * FROM rotation ORDER BY regex ASC").fetchall()
+                    return [list(c) for c in result]
+                except sqlite3.OperationalError:
+                    return []
 
     def create_mapping_table(self):
         """Create the mapping table."""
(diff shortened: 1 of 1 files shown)


Bouni:
I'll merge this as an emergency fix, I've tested it just roughly and hope it doesnt't break anything 🤞🏽

binary1230:
Cool, appreciate it. Yea sucks they broke the API.

If I get some more time this week I'll try and make this a little nicer, there's some situations where it's best to still delete parts.db by hand. We could probably warn the users a little more that it's expected that they get a 404 or something.

At least this fix will help with exports, which is still an awesome use of the plugin, even if the parts search stays bricked

Bouni:
@binary1230 I have a parts db from the 23rd of November, so pretty much the most recent state before they broke it.
I think I'll supply that from one of my webservers or put it into the repo. Not sure whats smarter ...
The database has ~480MB but zip compressed its just 58MB.

Any thoughts on this?

binary1230:
Yea, awesome.

I think shipping with it might not be a bad idea if it's only 58mb, especially since it looks like you already have some build scripts that could package it up into the zip that people download via kicad plug-in manager.

GitHub gets weird about files in the repo itself over.... I think 10mb unless you enablr LFS support which gets weird and sometimes expensive even for free public repos.

But if you hosted it on a webserver and had your build scripts download it when it makes the package, that's probably no issue

Then what you could maybe do is have the plugin try and download the URL, and if it 404s, display a message that the parts DB will be an older cached version, and that we're aware of an issue so for now they have to use the JLC site for the actual searching.

That's just kinda initial thoughts, if you have any other way in mind I'm sure it'll be fine. We all appreciate your efforts, sorry JLC Is wasting yours and everyone else's time with something they really should have foreseen and been proactive about

binary1230:
If you want another maintainer btw during this hecticness, happy to jump in, I'm pretty reliant on this plugin for a bunch of upcoming production runs I'm doing for a few PCBs, so plan on keeping an eye on stuff regardless.

PR #243, opened 2022-11-26
[cosmetic] UX: improve error msg if JLC URL fails to download

binary1230:
see #241 for impetus, URL is 404'ing.

this PR does nothing but just shows the URL so maintainers can better figure out what's going on from (what is sure to be) a bunch of incoming bug reports for the same thing in the near future. it's optional

see also: #242


Bouni:
Totally legit 👍🏽

PR #244, opened 2022-11-26
[dev only] add new standalone mode for startup

binary1230:
- motivation: I wanted to be able to debug the UI with PyCharm/vscode, but couldn't figure out how to get KiCad+python debugging to work.
- I wanted to set breakpoints to figure out what was going on with downloading and startup errors, which is hard to iterate on when you have to reload Kicad/PCBNew each time
- this is not the most stable thing in the world but gets the job done
- see __main__.py for instructions on how to launch this

the real work being done here is
- wrap all calls to GetBoard() with a class that can return a fake stub

this helped with creating #242, but is optional.


chmorgan:
@Bouni would recommend we close this in-lieu of the updated #394

PR #245, opened 2022-11-29
Replace JLCPCB CSV download with Fallback sqlite db download

Bouni:
As long as JLCPCB doesn't com up with a new way to download their parts catalog we download the last state from 2022-11-23 and use that.

### DIFF
--- a/library.py
+++ b/library.py
@@ -11,6 +11,7 @@ from enum import Enum
 from ntpath import join
 from pathlib import Path
 from threading import Thread
+import zipfile 
 
 import requests
 import wx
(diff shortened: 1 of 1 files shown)


Bouni:
@binary1230 Would you mind testing this PR? I replaced the CSV download with the download of the parts db from the 23rd, I guess thats the best workaround for the moment ...

binary1230:
Sure I'll try and get to it today

tmolteno:
Working here on Debian testing (see comment on Issue #241)

Bouni:
@tmolteno I'll make some minor changes as suggested by @binary1230 and then merge this.
Then you can install via PCM, no need for downloading the zip anymore.

binary1230:
Sorry I was slammed this whole week, sounds like it's working and looks good at a glance. Thanks so much for all of this.

PR #246, opened 2022-12-01
Make part selector assign multiple parts at once

bebidek:
Small improvement that fixes long freezes caused by running method `populate_footprint_list` for every assigned part.


Bouni:
Thanks for the PR!

PR #266, opened 2023-01-14
Use bounding box to determine component position center

computergeek1507:
JLC position parser is expecting the center of the component, Use bounding box to determine component center. Mostly effects through holes as their position is normally pin 1 not center

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -34,7 +34,7 @@ from pcbnew import (
     wxPoint,
 )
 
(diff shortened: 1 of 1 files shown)


Bouni:
Did a quick test with a 2x1 pin header, seems to work fine 👍🏽 

grafik

Bouni:
Thanks @computergeek1507 for your PR!

PR #298, opened 2023-03-24
added support for V7 Schematics with no id field

computergeek1507:


schematicexport.py
```diff
@@ -3,22 +3,34 @@
...
+    """This only works with KiCad V6/V7 files, if the format changes, this will probably break"""
 
     def __init__(self, parent):
         self.logger = logging.getLogger(__name__)
         self.parent = parent
 
     def load_schematic(self, paths):
-        for path in paths:
-            self._update_schematic(path)
+        if is_nightly(GetBuildVersion()):
```
radoslav006, line 23: Is function `is_nightly` the correct way of checking for KiCad 7? I am not a KiCad plugin developer, but I would think of something like this:
```
if GetBuildVersion().startswith('7'):
```

computergeek1507: currently "is_nightly" returns true for version 6.99, 7, or 7.99. It uses elsewhere for version 7 checking.

radoslav006: Thank you for the response and explanation, I see. I just think it will be less confusing, if the function name will be `is_kicad7` or something similar that describes what the function really does.

Bouni: The `is_nightly` function was introduced back when there was only a nightly V6.99 which contained breaking changes.
So the name is no longer meaningful, I think I should come up with a smarter way of handling the different versions.

sirtux:
Any updates on this PR? I'd really love to have this for kicad7!

Bouni:
Sorry, totally lost track on this one.
@sirtux Could you test this PR and let me know if it works for you in V7?

If that's the case I'll merge it and release a new version

sirtux:
Yeah, will test it, but can take some days - I'll put in on my todo for latest next weekend.

Thanks for your work!

sirtux:
Looks fine for me!

Bouni:
Thanks @computergeek1507 and @sirtux  for your help with this 👍🏽

synaption:
I have used this PR with v7 to get assembled boards.  It went smooth and they worked.  Thanks!

PR #307, opened 2023-05-01
add script to convert jlcparts db to plugin style db

markusdd:
Work in progress to resolve #304

Strategy: fetch jlcparts database which is generated daily, convert it to this plugins format to bring down size and enhance usability for the purposes of this plugin.
A github pipeline should then run daily, always serving the latest db, tehn just the download function of this plugin needs adaptation.

- [x] conversion script
  - [x] solve issue that in stock search is somehow not working although numbers are in
  - [x split zip in pure python so that it can be reassembled, Github limits to 100MB file size
- [x] pipeline
- [x] ~~adapt download function~~ will be done in a second PR

PR #310, opened 2023-05-06
implement download of new database source

markusdd:
this must be merged after #307

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -12,14 +12,29 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.11' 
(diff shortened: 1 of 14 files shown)


Bouni:
@markusdd could you please rebase this PR?

markusdd:
Hi, it is in fact 2 files: library.py and the unzip_parts.py.

I had to tweak the unzip_parts.py a little as there were issues with using Pathlib, hence the conflict.

I do not know why github lists all the commits again, they should not appear once this gets merged as they are already there.
I will try to rebase and see if they vanish.

markusdd:
yeah it just seems to list those commits here. When I try to rebase my branch it says up-to-date (which is true)
image

PR #315, opened 2023-05-23
Add setting to define which LCSC number source has priority

Bouni:
This lets the user decide if a LCSC number from the schematic overrules one in the database or should be ignored.

Screenshot 2023-05-23 115716

Screenshot 2023-05-23 115721


GhanGhan:
Hey
1) I would like to say thank you Bouni for making this plugin, it has made my life easier
2) I would recommend using radio buttons for this instead, as a new user this would have cleared up a lot of confusion for me
3) This option seems to not work consitently.  I replaced quite a few extended components with basic components, even though I clicked the checkbox so that the LCSC numbers would update from schematic, most but not all components were updated.  I even went as far as reinstalling the plugin.

PR #340, opened 2023-07-22
Cmm improvements - auto search, fts5 faster searching, a few other misc changes

chmorgan:
Note that there is a commit in this series that should be dropped if/when this is merged back to the main repo, was used so I could test this out locally.

I can split up a number of the earlier commits into separate PRs if you'd like, as well.


Bouni:
Please split the changes into separate PRs, that makes it a lot easier to review and test these changes!

PR #353, opened 2023-08-07
Fix icon colours for dark theme

serpent213:
Tested on MacOS 13.4 and current KiCAD 7.

Had to patch the scaling to 1.0 as well, BTW (Bouni#330).

Great plugin!


serpent213:
Screenshot 2023-08-07 at 23 58 11

Some eye candy for your entertainment.  🙂

Bouni:
Wasn't able to test this on windows as wx seems to not support dark mode here but looks like it can't cause to much trouble so I'll merge this one.

PR #365, opened 2023-08-30
Detect top/bottom layer correctly for flipped footprints

simpkins:
Check footprint.IsFlipped() when detecting top vs bottom layer placement to correctly handle flipped footprints.

Also fix an unhandled exception when logging about rotation corrections.

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -96,23 +96,24 @@ class Fabrication:
 
     def rotate(self, footprint, rotation, correction):
         """Calculate the actual correction"""
+        correction_int = int(correction)
         if get_is_top(footprint):
-            rotation = (rotation + int(correction)) % 360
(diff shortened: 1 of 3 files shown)


simpkins:
Actually, I think the issue here is a problem with some of the footprint definitions I was using, and not with the exporter.

I was using footprints for some reverse mount LEDs, and they footprints define the pads on the B_Cu layer.  However, Kicad appears to require that all footprints use the F_Cu layer as their primary layer.  (The Kicad BOARD_ITEM class supports defining the primary layer for items, but the FOOTPRINT class appears to have no way to set the primary layer to anything other than F_Cu.)

I think I simply need to fix the footprint libraries I was using to always define primary pads on the F_Cu layer rather than B_Cu.

Issue #374, opened 2023-09-20
rotation mapping automatically flip the rotation sign if the layer is top or bottom


gionag:
If i save a mapping for certain package on the top layer, when i use the same package on the opposite layer, also the sign of the rotation should change.

Should internally adapt the sign if the component is placed on top or bottom

Maybe is a recent change on jlcpcb portal

Bouni:
Can you give an example so that I can test it?

dzid26:
I see the same problem for my double sided board. 

But see the "sign" is already changed here:
https://github.com/Bouni/kicad-jlcpcb-tools/blob/b764df8f686480bc7b494119e9beed2e891d7c8c/fabrication.py#L94-L112
After removing this condition (i.e. always True), so far I've gotten correct results on both sides.

dzid26:
I was able to reproduce it with kicad library footprint.

Example footprint is TO-277A with C895438 - rotation 90deg - top:
![image](https://github.com/Bouni/kicad-jlcpcb-tools/assets/841061/a2bf02cf-abe4-44eb-b01e-01dda1363c2c)
Bottom:
![image](https://github.com/Bouni/kicad-jlcpcb-tools/assets/841061/b14c06be-2676-441b-a2f7-57f7e269e6a2)

z2amiller:
Fixed by #395

PR #395, opened 2024-01-10
Rotation correction bottom assembly

dzid26:
This fixes the problem that shows up relatively rare. #374 - That is it fixes rotation for parts that require rotation and are placed on top and bottom assembly.

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -91,24 +91,14 @@ class Fabrication:
 
     def rotate(self, footprint, rotation, correction):
         """Calculate the actual correction."""
-        if footprint.GetLayer() == 0:
-            rotation = (rotation + int(correction)) % 360
-            self.logger.info(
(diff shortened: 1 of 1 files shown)


Bouni:
Do you now if JLC or KiCAD has changed the beaviour of the rotation lately? I remember that I've implemented the logic to habdle top and bottom seperately because of this and now it nees to be removed again 🤔
Did you test with V7.99 or with V7?

dzid26:
The problem is not unique to 7.99. I had that thought and tested both.  
I am suspecting JLC changed the behavior because [similar issue](https://github.com/bennymeg/JLC-Plugin-for-KiCad/issues/101#issuecomment-1884504965)

Bouni:
I will try to find the time an dtest this on monday 🤞🏽

PR #398, opened 2024-01-14
Cmm direct - Direct calls and remove duplicate helper

chmorgan:
I'm not sure if the get_footprint_by_ref() is necessary for older versions of kicad but that could be a good reason to keep it.

The direct calls make it easier to determine which object is being referenced.

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -36,7 +36,7 @@ try:
 except ImportError:
     NO_DRILL_SHAPE = PCB_PLOT_PARAMS.NO_DRILL_SHAPE
 
-from .helpers import get_exclude_from_pos, get_footprint_by_ref
+from .helpers import get_exclude_from_pos
 
 
 class Fabrication:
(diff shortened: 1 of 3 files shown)


chmorgan:
Thoughts @Bouni?

Bouni:
Do you know when `FindFootprintByReference` was introduced? 
I think still a lot of people use V6 but the majority use V7.
If V6 compatibility breaks, I don't really care as V8 is about to be released, but V7 should be supported until V8 is stable.

chmorgan:
@Bouni its listed in the 6.0 docs here, https://docs.kicad.org/doxygen-python-6.0/classpcbnew_1_1BOARD.html#aedb712ddb30b1a91d2f9bb7d441c2fc8

I found a link for 5.x docs but it appears broken, it was an s3 bucket link that shows xml at this point, https://kicad-downloads.s3.cern.ch/doxygen-python/namespacepcbnew.html

So it appears to be valid in 6.x but maybe it wasn't present in 5.x.

Agree on supporting V6+ at this point.

chmorgan:
@Bouni ping on this one. I've got another interesting one queued after this lands.

PR #399, opened 2024-01-16
Add pseudo category 'All' and default to it.

DanielO:
This changes the default behaviour to how it was previously (ie search all categories) but still allows category selection.

### DIFF
--- a/library.py
+++ b/library.py
@@ -140,7 +140,7 @@ class Library:
         if "package" in parameters and parameters["package"] != "":
             p = parameters["package"]
             query_chunks.append(f'"Package" LIKE "{p}"')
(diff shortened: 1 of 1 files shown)


Bouni:
I'll look into all your PRs on Monday hopefully

DanielO:
> I'll look into all your PRs on Monday hopefully

Thanks, no hurry :)

Bouni:
@DanielO is this a MacOS only issue?

On Windows I can leave the categories empty and the search works across all of them:

grafik

But I think I remember someone said that on MacOS wx does not show the empty slot and selects the first category by default, right?

PR #401, opened 2024-01-16
Move stock

DanielO:
Move the stock column left a few spaces so it is visible by default.


DanielO:
Sorry about that, my patches were interdependent but I guess I operated git wrong :)

DanielO:
OK, I've munged it now

PR #402, opened 2024-01-16
Human price

DanielO:
Humanise the pricing, print in dollars or cents to 2 SF.


Bouni:
Same issues as with #401 please fix and rebase

DanielO:
I've fixed this up.
Note that if/when #401 goes in line 619 in partselector.py will have to change to `pricecol = 8`

Bouni:
I added a bit more space for better readability. It doesn't harm as the line is to long anyway and we should think about possible solutions for that, maybe some kind of tootip ...

DanielO:
Thanks - it is tricky.
IMO the "best" way would be to select the price based on how many you are using, but of course than would require knowing how many of this part are in the design and how many boards you are making..

Bouni:
Maybe we can show the details of the selected part in the right section 

grafik

and eliminate the Show part details button all together 🤔

Bouni:
I made a quick draft and I think that's the way to go:

grafik

Still needs a bit polish

DanielO:
Hmm I guess the main issue would be how long it takes to update the information when you click on an item - if it is not very fast then I fear it would be quite irritating to use.

PR #404, opened 2024-01-19
mainwindow.py - Select parts on the board in pcbnew when they are sel…

chmorgan:
…ected in the footprint_list list

Makes it easier to figure out which parts are being chosen, as a cross check for things like part size


Bouni:
This is a really cool one 😎

PR #426, opened 2024-03-05
fix #425 - part picture not loading in part details view

### LINKED ISSUE #425: Unknown Image Data Format Error when opening Part Details
Latest greatest KiCAD 8 and latest version of plugin.

I'm facing this when using the Part Details dialog.

grafik

I am unsure at this point if this has anything to do with KiCAD8 or if JLCPCB has played us here.

Suspicious code location where I think it goes wrong (the error must be form wxwidgets because it's nowhere in the plugin codebase):
https://github.com/Bouni/kicad-jlcpcb-tools/blob/20d6fd270286f2494fafde62ea4837a93fd34043/partdetails.py#L128

Any background info on this @Bouni? Happy to work on a possible fix, but I do not know what image format these were originally expected in and what wxwidgets can support out of the box.

markusdd:
fixes #425

issue is resolved by pretending we are a Browser, otherwise we get blocked. investigation is in related ticket

### DIFF
--- a/partdetails.py
+++ b/partdetails.py
@@ -127,7 +127,10 @@ class PartDetailsDialog(wx.Dialog):
 
     def get_scaled_bitmap(self, url, width, height):
         """Download a picture from a URL and convert it into a wx Bitmap."""
-        content = requests.get(url, timeout=10).content
+        headers = {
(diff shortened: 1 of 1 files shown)


Bouni:
Thnks for investigating this and for the PR!

PR #427, opened 2024-03-10
harmonize spelling and add assembly process param

markusdd:
attributes use capital spelling and also the existing hardcoded params were inconsistent. harmonized everything to capital spelling and added assembly process param (which is THT or SMT), which I think is actually pretty useful


markusdd:
mmh not sure why it lists the fix from the other day again, github quirk I guess.

Bouni:
Nice, thanks 👍🏽

PR #428, opened 2024-03-21
added common issue chapter

kyomawolf:
I found out that this may be an issue if you haven't worked with python on the system yet and this package is not installed


Bouni:
You might should add a note that the pip command has to be executed in a Kicad command prompt (on windows at least)

Bouni:
Will close this due to inactivity

PR #436, opened 2024-04-03
fix missing parenthesis

computergeek1507:


Bouni:
Thanks @computergeek1507 for fixing this so quickly! Just tested with V8.99 and it works perfectly 🚀

PR #437, opened 2024-04-04
Cmm jlcpcb script

chmorgan:
A few things found when rebuilding the fts5 db again.


chmorgan:
@Bouni ping on review of this one and the standalone changes

PR #443, opened 2024-04-20
Fix missing rotations display and sorting by Stock in the main list

gyohng:
Rotation column is empty regardless of the regex. This seems to be a display bug. This pull request brings it back.

### DIFF
--- a/library.py
+++ b/library.py
@@ -89,10 +89,10 @@ class Library:
             "Package",
             "Solder Joint",
             "Library Type",
+            "Stock",
             "Manufacturer",
             "Description",
             "Price",
-            "Stock",
         ]
         if self.order_by == order_by[n] and self.order_dir == "ASC":
             self.order_dir = "DESC"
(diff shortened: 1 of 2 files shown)


gyohng:
Additional explanations at the request of chmorgan:

It seems like Qt UI doesn't recognise the non-string type for a column, so it needs to be explicitly converted to string (the  [3b8595e](https://github.com/Bouni/kicad-jlcpcb-tools/pull/443/commits/3b8595ecd2d4b8fbe6860dee852f5b4e1772e29e) commit where str() is added around correction variable). Here's the result:

image

Sort order commit ([191cd95](https://github.com/Bouni/kicad-jlcpcb-tools/pull/443/commits/191cd95fb849711a433a4c674bd83383247f8691)) was different from the database column order list, so clicking the header would use a different column for sorting than the one displayed in the header, thus sorting wouldn't work as expected.

Issue #360, opened 2023-08-10
Database performance


serpent213:
The startup time can be greatly improved by creating an index on the LCSC part number:

```sql
CREATE UNIQUE INDEX parts_lcsc_part_index ON parts ('LCSC Part');
```

Before:

<img width="1440" alt="Screenshot 2023-08-10 at 20 42 38" src="https://github.com/Bouni/kicad-jlcpcb-tools/assets/106662/aaf26607-cb9b-483e-aa4c-f64c892231f1">

After:

<img width="1440" alt="Screenshot 2023-08-10 at 20 43 05" src="https://github.com/Bouni/kicad-jlcpcb-tools/assets/106662/3403e0b5-978b-4664-9ce1-def7135e7b4f">

(`populate_footprint_list()` from 2552ms to 137ms)

Another nice touch would be to open the database only once when the plugin starts up instead of whenever needed.

P.S.: I used [VizTracer](https://github.com/gaogaotiantian/viztracer) for the profiling:

```diff
--- a/plugin.py
+++ b/plugin.py
@@ -2,6 +2,7 @@
 
 import os
 
+from viztracer import VizTracer
 from pcbnew import ActionPlugin  # pylint: disable=import-error
 
 from .mainwindow import JLCPCBTools
@@ -25,6 +26,12 @@ class JLCPCBPlugin(ActionPlugin):
 
     def Run(self):
         """Overwrite Run"""
+        tracer = VizTracer()
+        tracer.start()
+
         dialog = JLCPCBTools(None)
         dialog.Center()
         dialog.Show()
+
+        tracer.stop()
+        tracer.save("/tmp/jlcpcb_profile.html")
```

Bouni:
As always, PRs are highly appreciated 😁

I simply lack the knowledge to improve stuff like that 😅

serpent213:
Some inspiration:

* You do a lot of string concatenation for query building, like
  ```python
  query += f' ORDER BY "{self.order_by}" COLLATE naturalsort {self.order_dir}'
  ```
  That might be bearable in this case (plugin for desktop use), but introduces SQL injection vulnerabilities.

  ![](https://imgs.xkcd.com/comics/exploits_of_a_mom.png)

  For proper escaping always use the SQL library's parameter handling (`cursor.execute(sql, [param_1, param_2])`).

[...]

PR #444, opened 2024-04-21
Create and use a separate table (parts_by_lcsc) to index FTS5 parts table by rowid to speed up.

gyohng:
The purpose of this PR is to increase the speed of any operation that involves list rebuilding, which is basically *any* operation, such as part selection, sorting, hiding/showing BOM parts, or even invoking the plugin window.

The database increase due to this is +3.3%, which I think is a reasonable trade-off until a better way is found to integrate it with FTS5 directly while maintaining the speed.

The problem is illustrated in this YouTube video (the principal part of the commit is initially commented out and then uncommented to show the difference in performance impact). 

[Youtube video with demonstrating the problem](https://www.youtube.com/watch?v=hYp7NG1VdmA)

The commit includes the following changes:

1. **Creation of a New Table (`parts_by_lcsc`)**: A new table is created in the database to hold mappings between part identifiers (`partsId`) and the `LCSC Part` numbers. This table is designed to optimise lookup times by indexing parts based on their `LCSC Part` number rather than through direct queries on the `parts` table, which seems to be slow.

2. **Modification in `get_part_details` Function**:
   - The method has been updated to first attempt fetching `partsId` values from the new `parts_by_lcsc` table using the provided `LCSC Part` numbers.
   - If the new method finds all required `partsId`, it then fetches the parts details directly using these IDs, which is much faster as it leverages the row IDs for direct access instead of scanning through text fields.
   - In case of any issues (e.g., the new table not having all the entries), it falls back to the original method of fetching data directly using the `LCSC Part` numbers from the `parts` table.

3. **Database Indexing in the `download` Method**:
   - A significant process added in the `download` method is the population of the `parts_by_lcsc` table and the removal of an old index if it exists.
   - It iterates through every row in the `parts` table, inserts or updates the mapping in the `parts_by_lcsc` table, and updates a progress indicator, which communicates the operation's progress to the user interface.
   - After filling up the new table, it creates an index on the `lcsc` column of the `parts_by_lcsc` table to speed up the lookups.
   - This indexing is expected to significantly reduce the query time by utilising a quicker, more efficient index.

4. **User Interface Updates**: The code updates a gauge control that reflects the progress of the database indexing operation. On a modern M1 MacBook Pro it finishes under 5 seconds and is significantly faster than the downloading.

### DIFF
--- a/library.py
+++ b/library.py
@@ -553,15 +553,14 @@ class Library:
             self.logger.debug("Indexing parts table...")
             wx.PostEvent(self.parent, UpdateGaugeEvent(value=0))
             with contextlib.closing(sqlite3.connect(self.partsdb_file)) as con:
+                con.execute('DROP TABLE parts_by_lcsc;')
                 con.execute('CREATE TABLE IF NOT EXISTS parts_by_lcsc (partsId INTEGER, lcsc TEXT);')
-                howMany = con.execute('SELECT COUNT(*) FROM parts').fetchone()[0]
                 con.execute('DROP INDEX IF EXISTS LCSCpartIdx;')
                 cur = con.execute('SELECT rowid, `LCSC Part` FROM parts')
+                indexedParts = cur.fetchall()
+                howMany = len(indexedParts)
                 progress = 0
-                for i in range(howMany):
-                    r = cur.fetchone()
-                    if r is None:
-                        break
+                for i, r in enumerate(indexedParts):
                     con.execute('INSERT OR REPLACE INTO parts_by_lcsc (partsId, lcsc) VALUES (?, ?)', (r[0], r[1]))
                     p = int(i / howMany * 100)
                     if p > progress:
@@ -573,6 +572,7 @@ class Library:
                 con.commit()
                 con.execute('CREATE INDEX IF NOT EXISTS LCSCpartIdx ON parts_by_lcsc(lcsc);')                
                 con.commit()
+                self.logger.debug("Indexing parts table done.")
 
             wx.PostEvent(self.parent, ResetGaugeEvent())
             end = time.time()
--- a/library.py
+++ b/library.py
@@ -383,6 +383,17 @@ class Library:
             numbers = ",".join([f'"{n}"' for n in lcsc])
 
             try:
+                rows = cur.execute(f'SELECT partsId FROM parts_by_lcsc where lcsc IN ({numbers})').fetchall()
+                if len(rows) == len(lcsc):
+                    numbers = ",".join([f'"{r[0]}"' for r in rows])
+                    return cur.execute(
+                        f'SELECT "LCSC Part", "Stock", "Library Type" FROM parts where rowid IN ({numbers})'
+                    ).fetchall()
+            except Exception as e:
+                self.logger.debug(f"{e}")
+                pass
+
+            try:
                 return cur.execute(
                     f'SELECT "LCSC Part", "Stock", "Library Type" FROM parts where "LCSC Part" IN ({numbers})'
                 ).fetchall()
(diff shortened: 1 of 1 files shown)


chmorgan:
Can we avoid rebuilding the list to reduce the impact of list rebuilding being expensive?

I get having a separate index table and considered such things. Imo it isn't worth the complexity if the cost of a few seconds is only when users alter the sort list of a component list with 100+ components.

gyohng:
> Can we avoid rebuilding the list to reduce the impact of list rebuilding being expensive?

I don't think it's possible if we want to keep the component list synchronised to the database. Even at a cost of a major refactor, you'll still need to do it every time the plugin starts, which is quite annoyingly slow right now.

In the view of the FTS5 change, which expanded the database four-fold (and if you download the second time, watch for the multi-gigabyte bak file), I think a db size increase of 3.3% and an extra SELECT is not such a big thing.

whmountains:
I second the observation that plugin startup has been very slow since the fts5 change.  The plugin also freezes for about 20 seconds whenever I select a part which sort of defeats the purpose of fts5 in the first place.  (I'm using MacOS and KiCAD 8 BTW)

I was just in the process of doing some profiling to troubleshoot the issue but it looks like @gyohng may have found it already.

chmorgan:
Agree any freezes should be eliminated or reduced.

For the freezing, what is the action that results in the freezing for you?
I’m assuming it’s something other than changing the parts sort order.

whmountains:
The slowness seems to come from the `populate_footprint_list` function.  I added some log statements to print the execution time whenever the function was called.  Here are some measurements I got in standalone mode:

Time taken in the `populate_footprint_list` function:
- Startup - 15.94s
- Click "Select part" from the part picker dialog - 13.74s
- Sort by stock in the main window - 13.40s.
```

chmorgan:
That helps. I'll take a look to see what might be going on there. We know
the fts5 stuff is a lot of overhead (in exchange for the free text
searching), there may be some straightforward wins to cut that time down on
those steps you mention taking a long time.

In terms of performance, I wrote a rust program that does profiling
against the old and new databases, and various versions in between,
relative to search queries, and it was helpful. I'll do the same here for
these paths.

whmountains:
I think #360 is related.  The issue was actually fixed in 5eb59e2, but the issue never got closed.  We lost this index with the move to FTS5.

gyohng:
> I think #360 is related. The issue was actually fixed in [5eb59e2](https://github.com/Bouni/kicad-jlcpcb-tools/commit/5eb59e215d8a304a47f543f35659f3eeb852c9e8), with the comment "massive startup time improvement" but it never got closed. We lost this index with the move to FTS5, which caused startup time to once again take a lot longer.
 
Hi! :)  I am actually a fan of your previously submitted PR with LIKE optimisation, but now FTS5 took over...

This PR is essentially a replica of 5eb59e2. The main problem with 5eb59e2 (indexing by `LCSC Part`) now is that there's no way to reintroduce it with FTS5 - the FTS5's own index isn't quite as efficient, and sqlite won't accept index creation for an FTS table.

I went even as far as creating an index on the internal `parts_content` table for the FTS structure. If I had to choose, I'd probably keep the FTS5 part separately just for the part search and otherwise resort to the old table with index approach for everything else, even despite the data growth.

whmountains:
Hi @gyohng!  I thought I already replied but I can't find the comment so I'll reply again.

I think this comes down to workflow.  You and I make heavy use of parametric search, so #439 was really all we needed and FTS5 has a high bar to meet.  @chmorgan and probably a lot of other people prefer a single search-box interface hence the project going in that direction.

But I think we all agree that the UI freezing for 15 seconds or more every time a part is selected is not acceptable to anyone's workflow.  You may be interested in the discussion under #451, where we are trying to figure out where the slowdown is coming from.  And #450 which is a totally diffferent idea for a parts database.

Bouni:
I just tried this PR but it always stops at "DEBUG - download -  Indexing parts table..."
The progress bar is not filling up so I guess something is not working as expected.
@gyohng Any ideas why?

gyohng:
> I just tried this PR but it always stops at "DEBUG - download - Indexing parts table..." The progress bar is not filling up so I guess something is not working as expected. @gyohng Any ideas why?

It used to work fine, I will check with the latest pull. By the way, I noted that your logging is buffered and not flushed, so you normally won't see things until several lines (sometimes minutes) afterwards.

gyohng:
Please check again. 'IF EXISTS' clause was missing from DROP TABLE. I could never run in to it at my end, because this table was already indexed with a previous version. Now it should work fine.

gyohng:
I added another commit to make linter happy with this PR.

Oscilllator:
I didn't see this pull request prior to taking a look at the performance myself, but I was able to make get_part_details 100x faster without too much trouble in https://github.com/Bouni/kicad-jlcpcb-tools/pull/503. From talking with my good friend ChatGPT it seems that many of the SQL queries seem to not take into account the FTS5 nature of the table.

chmorgan:
@Oscilllator speeding the query without the additional table is a simpler approach.

gyohng:
@chmorgan if the other one works the same fast, let's close then?

chmorgan:
@gyohng it might not be as fast but it does appear like its faster (note that I'm back using the plugin to release some boards but I haven't tested it). Your call on closing this one if you think its been solved, or keeping it open if you think this is a better approach.

Bouni:
I'll close this as its no longer an issue

PR #445, opened 2024-04-21
Fix missing rotation corrections display in the main view

gyohng:
This patch corrects an issue where the rotation data was not being displayed correctly in the UI due to type recognition problems with the wx framework. Previously, certain rotation correction values were not converted to strings before adding to a wx list, leading to them being ignored by the GUI framework, showing no data and not reporting any errors.

The fix involves explicit string conversion of the rotation data in the populate_footprint_list method. Specifically, when correction rules are applied based on part names or packages, the corrected rotation values (part[8]) are now converted to strings (str(correction)) before being set. This change ensures that the wx framework recognises and displays the data correctly.

Please refer to this picture:
image

### DIFF
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -616,13 +616,13 @@ class JLCPCBTools(wx.Dialog):
             # First check if the part name mathes
             for regex, correction in corrections:
                 if re.search(regex, str(part[1])):
-                    part[8] = correction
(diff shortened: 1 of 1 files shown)


chmorgan (approved):
Rational makes sense, same approach is also performed on line 588.

PR #446, opened 2024-04-21
Fix inconsistency between column ordering in the GUI and the database reference.

gyohng:
This patch addresses an issue where the column ordering in the GUI did not match the order used in database queries. This led to improper sorting functionality within the application in the part selection dialogue. The inconsistency was observed when a header was clicked to sort by a specific field, where the list would be sorted using a wrong field.

The position of the "Stock" field in the ordering array was adjusted to precede the `Manufacturer`, `Description`, and `Price` fields in the way it is in the GUI.

### DIFF
--- a/library.py
+++ b/library.py
@@ -89,10 +89,10 @@ class Library:
             "Package",
             "Solder Joint",
             "Library Type",
+            "Stock",
             "Manufacturer",
             "Description",
             "Price",
-            "Stock",
         ]
         if self.order_by == order_by[n] and self.order_dir == "ASC":
             self.order_dir = "DESC"


library.py
```diff
@@ -89,10 +89,10 @@ def set_order_by(self, n):
             "Package",
             "Solder Joint",
             "Library Type",
+            "Stock",
```
chmorgan, line 92: commit log mentions this has to match the database order, where is the defined? can we use the same list in both places to avoid having the two lists get out of sync?

gyohng: This, what you're looking at *is* the database column reference order.

The GUI list is in partselector.py line 368 and on. In theory, one could use a single source for it, but it's going to be a much more involved refactoring:

image

Plugging it right now is more sensible than leaving it unfixed until someone (you) finds enough time to revamp partselector.py, for which there could be a TODO task to fix it in the future.

gyohng: I need to get back to my projects now, cannot spend any more time on this. Please feel free to get inspired by this:

```py
# Defining the structure of each column - put this inside library and assign to a member
sel_columns = [
    {"name": "LCSC",         "db_name": "LCSC Part",     "width": 60,  "align": wx.ALIGN_LEFT,   "ellipsize": wx.ELLIPSIZE_NONE},
    {"name": "MFR Number",   "db_name": "MFR.Part",      "width": 140, "align": wx.ALIGN_LEFT,   "ellipsize": wx.ELLIPSIZE_NONE},
    {"name": "Package",      "db_name": "Package",       "width": 100, "align": wx.ALIGN_LEFT,   "ellipsize": wx.ELLIPSIZE_NONE},
    {"name": "Pins",         "db_name": "Pins",          "width": 40,  "align": wx.ALIGN_CENTER, "ellipsize": None},
    {"name": "Type",         "db_name": "Library Type",  "width": 50,  "align": wx.ALIGN_LEFT,   "ellipsize": wx.ELLIPSIZE_NONE},
    {"name": "Stock",        "db_name": "Stock",         "width": 50,  "align": wx.ALIGN_CENTER, "ellipsize": wx.ELLIPSIZE_NONE},
    {"name": "Manufacturer", "db_name": "Manufacturer",  "width": 100, "align": wx.ALIGN_LEFT,   "ellipsize": wx.ELLIPSIZE_NONE},
    {"name": "Description",  "db_name": "Description",   "width": 300, "align": wx.ALIGN_LEFT,   "ellipsize": wx.ELLIPSIZE_NONE},
    {"name": "Price",        "db_name": "Price",         "width": 100, "align": wx.ALIGN_LEFT,   "ellipsize": wx.ELLIPSIZE_NONE}
]
......

    for col in parent.library.sel_columns:
        col_width = int(parent.scale_factor * col["width"])
        renderer = self.part_list.AppendTextColumn(
            col["name"],
            mode=wx.dataview.DATAVIEW_CELL_INERT,
            width=col_width,
            align=col["align"],
            flags=wx.dataview.DATAVIEW_COL_RESIZABLE
        ).GetRenderer()

        if "ellipsize" in col and col["ellipsize"] is not None:
            renderer.EnableEllipsize(col["ellipsize"])
```

chmorgan: @gyohng yep, I'll pick it up

gyohng:
Rebased to the main branch. Please consider merging.

Bouni:
I'll close this as its already fixed in main. I totally overlooked your PR while I fixed that, sorry 😓

Issue #448, opened 2024-04-23
Crash during DB download on MacOS 


velolala:
Hi there, this plugin looks awesome, unfortunately I could not try it out yet, because of the following issue:

**Bug/Reproduction**
Every single time I open the plugin, it starts to download the zip parts of the parts DB, but then KiCAD crashes before completion. I monitored the `$plugin/jlcpcb` folder, and the furthest the download ever got was parts 001-007 complete (`80000000 bytes`), but part 008 still at `0 bytes`. Unfortunately I could not get a readable exception, but if you can help me to run the download step manually, I'm happy to debug further!

**Version information**
```
OS: MacOS 13.6.6 (22G630)

Plugin: 2024.04.03

Application: KiCad arm64 on arm64
Version: 8.0.1, release build
Libraries:
    wxWidgets 3.2.4
    FreeType 2.13.2
    HarfBuzz 8.3.0
    FontConfig 2.15.0
    libcurl/8.4.0 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.51.0
Platform: macOS Ventura Version 13.6.6 (Build 22G630), 64 bit, Little endian, wxMac
Build Info:
    Date: Mar 14 2024 13:30:33
    wxWidgets: 3.2.4 (wchar_t,wx containers)
    Boost: 1.84.0
    OCC: 7.7.2
    Curl: 7.87.0
    ngspice: 42
    Compiler: Clang 14.0.3 with C++ ABI 1002
Build settings:
```

ahalekelly:
I updated to the latest version using the KiCAD plugin manager and now I am also having this issue

velolala:
Great to hear from you! I actually found some more info: it seems there is some (b)locking issue, that triggers an OS watchdog. I started `kicad` from commandline and got this:

```
% /Applications/KiCad/KiCad.app/Contents/MacOS/kicad
Fontconfig error: Cannot load default config file: No such file: (null)
2024.04.24 10:05:20 - DEBUG - download -  Parts db is split into 8 parts. Proceeding to download...
2024.04.24 10:05:20 - DEBUG - download -  Removing any spurios old zip part files...
2024.04.24 10:05:20 - DEBUG - download -  Removing /Users/myuser/Documents/KiCad/8.0/3rdparty/plugins/com_github_bouni_kicad-jlcpcb-tools/jlcpcb/parts-fts5.db.zip.001.
[...]

PR #455, opened 2024-04-26
unzip_parts.py - Reduce working memory size by reading and writing sm…

chmorgan:
…aller chunks into memory vs. the whole zip file parts

Reduces the data read into ram from 76MB (the size of each zip file at present) to 1MB.


whmountains:
For posterity: see discussion in #448 for details of the logic behind commit cf2de77.

Bouni:
From 107s down to 47s, nice 😎

PR #477, opened 2024-05-23
Added rotation by reference

maly95:
Sometimes there is need to add rotation for only one footprint by reference name (every others by package or name matches correctly). Added possibility to set this rotation by reference (\b{name}\b regex) and checking regex also for footprint references.


gyohng:
Is the rotation database per-project or global?

chmorgan:
@dhoepfl would you mind taking a look at this?

@maly95 could you look at the work @dhoepfl has done in this area that has recently landed on main?

If it makes sense to continue this PR please comment here and we can do so.

PR #478, opened 2024-05-27
Cmm crash fix

chmorgan:


KnightHill:
Installed the fix and downloaded all jclpbc parts. Everything seems to be back to normal now (kicad 8.0.2 under Manjaro). Thank you so much for fixing the issue!

Bouni (approved):

Bouni:
Thanks @chmorgan for digging into this.

The log function needs an overhaul anyway. I've stolen it from KiBuzzard plugin and it has sometimes strange behavior. I would really like it if we could write to a log file as well (maybe that can be enabled in the settings!?)
What do you think?

PR #480, opened 2024-06-07
Panel support

### LINKED ISSUE #440: kikit panel CPL file  all over the place
I think this is a regression of https://github.com/Bouni/kicad-jlcpcb-tools/issues/155. I submitted a 3x3 board and had my CPL rejected. It seems that its randomly selecting a reference designation for each part, leading to the placement being all over the panel rather than on one unit.

Bouni:
This should fix #440

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -258,9 +258,10 @@ class Fabrication:
             writer.writerow(
                 ["Designator", "Val", "Package", "Mid X", "Mid Y", "Rotation", "Layer"]
             )
-            for part in self.parent.store.read_pos_parts():
-                fp = self.board.FindFootprintByReference(part[0])
-                if get_exclude_from_pos(fp):
+            footprints = sorted(self.board.Footprints(), key = lambda x: x.GetReference())
+            for fp in footprints:
+                part = self.parent.store.get_part(fp.GetReference())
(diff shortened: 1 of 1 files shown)


chmorgan (approved):
Only comment I'd make is that the commit subject and body for the first commit doesn't help with understanding the change. For kit kit are there multiple parts with the same refdes? In the fixed case clearly it's iterating over all footprints via self.board.Footprints() vs. what might not be all footprints(?) using read_pos_parts()?

Bouni:
>For kit kit are there multiple parts with the same refdes?

Exactly, KiKit takes a board an multiplies it so that it can be manufactured as a panel.
That means on a 2x2 panel we have 4x C1 for example.

The problem is the line after the `read_pos_parts`:

https://github.com/Bouni/kicad-jlcpcb-tools/blob/main/fabrication.py#L262

`board.FindFootprintByReference` gives a random position for a given ref, that way we end up with jus one part in the POS / BOM file.

I did it the other way, iterating over all parts I get from the board and look that up in the store. 

I hope that makes sense 😅

chmorgan:
If you could put that info into the commit message imo that would help the
next person follow the change.

This is what I had suspected the change was. That we were iterating over
all parts, grouping and matching by refdes for the component info.

PR #487, opened 2024-06-21
Use dict_factory in store for better readability

Bouni:
I always struggled with al those index numbers scatterd all around the code base and for a long time wanted to improve that.

So I started with the store and introduced a [row_factory](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.row_factory) that returns a dict instead of a list which improves readability throughout the code base a lot.

For example

`if self.hide_bom_parts and part[6]:`

becomes

`if self.hide_bom_parts and part["exclude_from_bom"]:`

### DIFF
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -780,7 +780,7 @@ class JLCPCBTools(wx.Dialog):
             bom = toggle_exclude_from_bom(fp)
             pos = toggle_exclude_from_pos(fp)
             self.store.set_bom(ref, int(bom))
-            self.store.set_pos(ref, pos)
+            self.store.set_pos(ref, int(pos))
             self.footprint_list.SetValue(
                 GetListIcon(bom, self.scale_factor), row, Column.BOM
             )
@@ -813,7 +813,7 @@ class JLCPCBTools(wx.Dialog):
             board = self.pcbnew.GetBoard()
             fp = board.FindFootprintByReference(ref)
             pos = toggle_exclude_from_pos(fp)
-            self.store.set_pos(ref, pos)
+            self.store.set_pos(ref, int(pos))
             self.footprint_list.SetValue(
                 GetListIcon(pos, self.scale_factor), row, Column.POS
             )
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -554,7 +554,7 @@ class JLCPCBTools(wx.Dialog):
         """Assign a selected LCSC number to parts."""
         for reference in e.references:
             self.store.set_lcsc(reference, e.lcsc)
-            self.store.set_stock(reference, e.stock)
+            self.store.set_stock(reference, int(e.stock))
         self.populate_footprint_list()
 
     def display_message(self, e):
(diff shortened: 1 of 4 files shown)


store.py
```diff
@@ -85,13 +86,11 @@ def read_all(self):
         """Read all parts from the database."""
```
chmorgan, line 86: should we annotate this with a '-> dict' hint?

store.py
```diff
@@ -183,57 +184,77 @@ def set_lcsc(self, ref, value):
...
+            db_part = self.get_part(board_part["reference"])
             # if part is not in the database yet, create it
-            if not dbpart:
+            if not db_part:
                 self.logger.debug(
                     "Part %s does not exist in the database and will be created from the board.",
-                    part[0],
+                    board_part["reference"],
                 )
+<<<<<<< HEAD
```
chmorgan: looks like a merge conflict snuck in here

store.py
```diff
@@ -122,17 +122,17 @@ def create_part(self, part):
...
-                    "UPDATE part_info set value = ?, footprint = ?, lcsc = ?, exclude_from_bom = ?, exclude_from_pos = ? WHERE reference = ?",
-                    part[1:] + part[0:1],
-                )
-            else:
-                cur.execute(
-                    "UPDATE part_info set value = ?, footprint = ?, exclude_from_bom = ?, exclude_from_pos = ? WHERE reference = ?",
-                    part[1:] + part[0:1],
-                )
-
+            cur.execute(
```
chmorgan, line 116: this is WAY simpler than the original version.

mainwindow.py
```diff
@@ -779,7 +779,7 @@ def toggle_bom_pos(self, *_):
             fp = board.FindFootprintByReference(ref)
             bom = toggle_exclude_from_bom(fp)
             pos = toggle_exclude_from_pos(fp)
-            self.store.set_bom(ref, bom)
+            self.store.set_bom(ref, int(bom))
```
chmorgan, line 782: what was up with the necessity to cast here but not previously? (No mention in the commit message of the reason why)

store.py
```diff
@@ -178,13 +178,6 @@ def update_from_board(self):
                     "Part %s does not exist in the database and will be created from the board.",
                     board_part["reference"],
                 )
-<<<<<<< HEAD
```
chmorgan: you could do a rebase and fixup this commit into the earlier one where the merge conflict was introduced, if you are comfortable with using git like that.

chmorgan (commented):
A handful of minor comments, this is a huge cleanup. I like the idea of using the columns directly from the database to generate the dict, that way if the order does ever change none of the code will have to.

store.py
Bouni: Absolutely, will add that in a minute

store.py
Bouni: Thats the point of this PR, I should have used a dict factory from the beginning, but wasn't aware that this is even possible ...

mainwindow.py
Bouni: I don't remember but I guess that sqlite doesn't like to get a "1" for a boolean value. Previously the casting was done in th e set_bom function, in other places it was done like this. So I decided to make the cast always in the same place.

chmorgan (approved):

Bouni:
Thanks @chmorgan for the review 👍🏽

PR #496, opened 2024-07-04
WIP: Show the price for the number of selected parts

Bouni:
This is an attempt to make the messy price column more useful.

We now go through the price ranges and check for the number of selected parts (doable via the alike button fr example) in whch price range we are. Then return that as the price:

grafik

### DIFF
--- a/partselector.py
+++ b/partselector.py
@@ -643,7 +643,7 @@ class PartSelectorDialog(wx.Dialog):
         self.search(None)
 
     def get_price(self, quantity, prices):
-        """Find the prce for the number of selected parts accordning to the price ranges."""
+        """Find the price for the number of selected parts accordning to the price ranges."""
         price_ranges = prices.split(",")
         min_quantity = int(price_ranges[0].split("-")[0])
         if quantity <= min_quantity:
--- a/partselector.py
+++ b/partselector.py
@@ -656,7 +656,6 @@ class PartSelectorDialog(wx.Dialog):
                 return float(price)
             lower = int(lower)
             upper = int(upper)
-            self.logger.debug(upper)
             if lower <= quantity < upper:
                 return float(price)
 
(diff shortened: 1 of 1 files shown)


Bouni:
I need to check better what 20-180 for example means. 

Does that include 20 but not 180? 
Why is there a gap to the next price range (20-180, 200-1180 and so on). 
Are the prices always in the right order so that the first range automatically contains the minimum quantity for the price?

partselector.py
```diff
@@ -600,6 +600,24 @@ def update_subcategories(self, *_):
         # search now that categories might have changed
         self.search(None)
 
+    def get_price(self, quantity, prices):
+        """Find the prce for the number of selected parts accordning to the price ranges."""
```
chmorgan, line 604: typo according

partselector.py
```diff
@@ -600,6 +600,24 @@ def update_subcategories(self, *_):
         # search now that categories might have changed
         self.search(None)
 
+    def get_price(self, quantity, prices):
```
chmorgan, line 603: could I suggest that we round to say three decimal points? like if price > 0.001 then price = round(ceiling(price), 3)?

Two arguments for dropping precision:
- This provides the per-board cost and if it's like $0.001 dollars total then it might as well be zero.

- If the cost of these passives matters for volume we'd need some way to let the user enter total board count per year or something to give them a total part cost. I usually manage my bom in a separate spreadsheet where I can calculate 1k pricing and a total burdened cost etc with margin etc.

chmorgan: Oh and it would make it easier to read that column, the small price parts are something I ignore so having a column that mostly lines up like 0.001 values would make it easier to skim over them.

Bouni:
I added the sum as well, which makes sense in my opinion:

grafik

Bouni:
But I found another issue, every part has a minimum price:

grafik

That piece of information is not in the parts database and if we want to show the correct price we would have to do an API call for every row in the results.
That would lead to hundrets if not thousends of requests which I guess is not what JLC / LCSC want us to do.

So we will not be able to show the real price for most parts 🤷🏽‍♂️

PR #505, opened 2024-07-28
Adds "LCSC Part" as an LCSC property; a "Description" column is shown to the user.

gonzalop:
-Adds "LCSC Part" as an LCSC property.
-Adds a "Description" column to the footprint list.
-Adds a new setting for the schematic property name.
-Hooks up the property name setting to the schematic export.

### DIFF
--- a/library.py
+++ b/library.py
@@ -387,12 +387,12 @@ class Library:
             con.row_factory = dict_factory
             cur = con.cursor()
             results = []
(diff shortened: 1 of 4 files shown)


chmorgan:
@gonzalop possible to include some screenshots showing what this looks like and comments about why you've made the change?

gonzalop:
This is how it looks like:
Screenshot from 2024-07-31 13-22-53

As for the comments, can you be more specific? One of the commits has the reason to add 'LCSC Part' (easyeda2kicad uses that name when importing footprints and 3dmodels, and this is also the LCSC database column name)

Bouni:
Obsolete due to #551

Issue #66, opened 2021-10-29
Very inconvenient to search for certain values like '100Ohm 0402 resistor'


t123yh:
When you search for '100 0402' or '100 0402 resistor', thousands of results appearing in the result box. 
![Screenshot_20211029_155648](https://user-images.githubusercontent.com/8972552/139397507-e7d5a621-2521-4f06-b650-53987538a188.png)

However, you can get the desired result if you search with '100Ω'.
 
Considering the fact that most people don't have the Ω symbol on their keyboard, I think it would be nice to have some form of auto-conversion in the search box, like converting `ohm` to `Ω`.

Bouni:
Maybe we can create some kind of smart filter here.

Like writing 100 pkg:0402 will search for 0402 in the packages, mfr:xxx could be the same for manufacturers.

What do you think of this approach?

t123yh:
> Maybe we can create some kind of smart filter here.
> 
> Like writing 100 pkg:0402 will search for 0402 in the packages, mfr:xxx could be the same for manufacturers.
> 
> What do you think of this approach?

Yes, I think it will be nice. Also, I think search box content can possibly be automatically filled from the component value.

Bouni:
Thats a nice idea, I like that!

yaqwsx:
If you want, you can use preprocessed libraries from my project: JlcParts (https://github.com/yaqwsx/jlcparts).

All basic values like resistance, capacitance, inductance, voltage, current, etc. are normalized to SI units which allows you to, e.g., sort properly by values. The libraries are JSON files per category and are served via GH-Pages.

Bouni:
@yaqwsx Thank you for this kind offer, I have to look into it if and how I could use your librarys.
I planned to set up a web api that makes me able to query the JLC parts in a nice way for quite a long time.
That also makes us able to query the data even when the JLC csv file is 404 again as it happened a few time the last couple of days!
I've seen that you have trouble with firefox and your plugin, this API might solve your browser issue as well!?
What do you think of a joint effort to setup such an API and use it in both projects?
I would pay for the infrastructure, I guess a [small virtual server (CX21)](https://www.hetzner.com/cloud) is sufficiant for such a service.
Let me know what you think 😃

yaqwsx:
[...]

PR #508, opened 2024-07-30
Add buttons to easy add  Ω and µ to the search string. Fix for #66

Bouni:
As requested ages ago in #66 I added two button which allow to easily add the Omega and the Micro symbol to the search string.

grafik


Bouni:
🙄 I just realized that JCL uses `uH` instead of `µH` in their description ....

Bouni:
Ok, but they use `µF` and not `uF`, so all good 😅

chmorgan (approved):

PR #509, opened 2024-07-31
Refactor LCSC API into seperate class

Bouni:
With this change the LCSC API can be used in other parts of the code and its a lot cleaner overall.


chmorgan (approved):
Looks good. I'd recommend squashing the commits together into a single commit during the merge for git history cleanliness but otherwise looks good.

Bouni:
I've never used squash merge before but now that you pointed it out it is pretty useful 😅

PR #517, opened 2024-08-06
WIP: Fix sorting of the footprint list

Bouni:
The sorting is broken, this is an attempt to fix it


Bouni:
This is obsolete as we no longer sort using the database

PR #519, opened 2024-08-07
Improve part list

Bouni:
This PR will replace the ListCtrl used for the main parts list with a DataViewCtrl and a PyDataViewModel.

We will be able to sort the list without having to query the database nd re-populate the list every time.


Bouni:
I guess I sucessfully migrated everything to the new list 🥳 

As I said, a heavy change, so please test and let me know if you find errors!

I realized that we have some minor issues which are not introduced by this PR, but I guess I'll fix them later.

Bouni:
Here's a screenshot, not much visual difference:

grafik

But some minor details that I find quite handy:

- Zebra style rows for better visibility
- Small Arrows on the column headers that show the sorting direction
- I decided to use an Icon to show the layer (TOP or BOT)

datamodel.py
```diff
@@ -0,0 +1,164 @@
...
+    def natural_sort_key(s):
+        """Return a tuple that can be used for natural sorting."""
+        return [
+            int(text) if text.isdigit() else text.lower()
+            for text in re.split("([0-9]+)", s)
+        ]
+
+    def GetColumnCount(self):
+        """Get number of columns."""
+        return 10
```
chmorgan, line 60: can we return the columns used in GetColumnType() and use len() on that to avoid hardcoding this value?

mainwindow.py
```diff
@@ -355,106 +355,154 @@ def __init__(self, parent, kicad_provider=KicadProvider()):
...
+        footprint.SetSortable(True)
+        lcsc.SetSortable(True)
+        type.SetSortable(True)
+        stock.SetSortable(True)
+        bom.SetSortable(True)
+        pos.SetSortable(False)
+        rotation.SetSortable(True)
+        side.SetSortable(True)
+
+        # self.footprint_list = wx.dataview.DataViewListCtrl(
```
chmorgan: should we drop the commented out lines?

mainwindow.py
```diff
@@ -576,29 +626,31 @@ def get_correction(self, part: dict, corrections: list) -> str:
...
-                "",
-                loadIconScaled(
-                    "mdi-close-color.png",
-                    self.scale_factor,
-                ),
-            ),
-        }
+        self.partlist_data_model.RemoveAll()
+
+        # icons = {
```
chmorgan: and here?

mainwindow.py
```diff
@@ -405,91 +409,6 @@ def __init__(self, parent, kicad_provider=KicadProvider()):
         rotation.SetSortable(True)
         side.SetSortable(True)
```
chmorgan, line 414: ahh k, could always rebase and fixup the previous change with this one if that makes sense with your git-fu

chmorgan (approved):
Looks like we've got a few hardcoded column indexes still at the end there but those are legacy. Overall PR looks clean, no complaints, and is moving in a good direction.

datamodel.py
Bouni: Done!

mainwindow.py
Bouni: No idea what you mean by that 😅

PR #521, opened 2024-08-08
Eliminate all dead code

Bouni:
I was aware that there are a lot of unused fuctions and vataibles throughout the code.

So I used [deadcode](https://github.com/albertas/deadcode) to find all unused functions and remove them.

In places where we overload functions I added `# noqa` comments

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -28,3 +28,16 @@ jobs:
           version: 0.5.6
           args: check 
           src: "."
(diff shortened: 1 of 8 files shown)


Bouni:
I decided to add deadcode to the CI so no more dead code can occure.
There is a problem with the pre-commit-hook for it, so I commented that for now and opened an issue (https://github.com/albertas/deadcode/issues/16)

PR #547, opened 2024-10-09
Fix LCSC numbers appearing on silkscreen

### LINKED ISSUE #546: All LCSC numbers appear on silkscreen
I don't know if this is a feature but after assigning part numbers they all show up on the silkscreen which looks like this:

image

image

I tried searching online how to bulk hide these fields but found no way (other that scripting).

Using Edit -> Edit Text & Graphics Properties does not work on these fields.
If I select ALL filters in that menu and set Visible -> False, all graphics disappear except these LCSC numbers.

Is this something that everyone deals with or am I the only one with this problem?

I'm using kicad 8.0.5 on ubuntu 22.

whmountains:
Fixes #546

The Python scripting API is a bit awkward.  It would be nice if `pcbnew.FOOTPRINT.SetField` returned the resulting field so we could call `SetVisible(False)` on it.

I looked for ways to bypass the call to `GetFieldByName()`.  The [source code for SetField](https://docs.kicad.org/doxygen-python-8.0/pcbnew_8py_source.html#l19337) has some ideas.  But it wasn't working quite right and I would have to modify the function signature to take `pcbnew` as an argument for use in standalone mode.  I think we would also have to modify our standalone impl to add a couple of extra stub methods.  

Thanks @Carbon225 for the example code showing the problem and a prototype solution.

### DIFF
--- a/helpers.py
+++ b/helpers.py
@@ -133,14 +133,19 @@ def get_lcsc_value(fp):
 
 def set_lcsc_value(fp, lcsc: str):
     """Set an lcsc number to the first matching propertie of the footprint, use LCSC as property name if not found."""
-    lcsc_field = "LCSC"
+    lcsc_field = None
     for field in fp.GetFields():
         if re.match(r"lcsc|jlc", field.GetName(), re.IGNORECASE) and re.match(
             r"^C\d+$", field.GetText()
         ):
-            lcsc_field = field.GetName()
-    fp.SetField(lcsc_field, lcsc)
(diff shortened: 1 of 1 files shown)


whmountains:
The original push would unconditionally set the LCSC field to be hidden, but I realized that was a bit unclean.  What if a user had manually set the field to be visible?  We wouldn't want the plugin to make it hidden in that case.  So I force-pushed a new implementation which only sets the field to visible if it did not already exist.

PR #551, opened 2024-11-10
Show component values in list (+ other minor conveniences)

wrs:
There are several commits in this PR, sorry. However, they are all independent, so you can apply whichever you approve of.

They're all small conveniences except the "**Show parameters derived from LCSC description**" commit, which addresses the following pain points:

1. LCSC often buries the most important data (like the resistance of a resistor) at the end of the description field. For example: "50mW Thick Film Resistors ±200ppm/℃ ±1% 234kΩ". The "234kΩ" is past the point where the part details dialog truncates the description. (And that dialog is not resizable, which is one of the minor fixes.)

2. When choosing a part, the values aren't visible in the list. If you search for "5kΩ" you'll also get "25kΩ", "1.5kΩ", etc. and you can't see which is which without opening the details dialog.

3. When double-checking the assignments, you can't just look down the footprint list to be sure the values are sane, because you can't see the values.

This commit tries to address this by implementing heuristics to pull the important data out of the LCSC description field and show it in the footprint and parts lists. For example:
image
image
You can easily scan the list to make sure the values and footprints match.

In part selection:
image
It's easy to tell the 5.1k from the 51k from the 1k.

This (in addition to implementing double-click in another commit) has vastly accelerated my workflow.

There's one other minor change to note: I altered the default setting of `lcsc_priority`. With the current setting, my experience as a new user was that I put in 15 minutes of work assigning parts, closed the dialog, and immediately lost all my work because I didn't know to click the "export" button. It seems safer to default the other way.

### DIFF
--- a/datamodel.py
+++ b/datamodel.py
@@ -5,8 +5,6 @@ import re
 
 import wx.dataview as dv
 
-from .derive_params import params_for_part
-
 from .helpers import loadIconScaled
 
 
--- a/datamodel.py
+++ b/datamodel.py
@@ -5,6 +5,8 @@ import re
 
 import wx.dataview as dv
 
+from .derive_params import params_for_part
+
 from .helpers import loadIconScaled
 
 
@@ -25,6 +27,7 @@ class PartListDataModel(dv.PyDataViewModel):
             "POS_COL": 7,
             "ROT_COL": 8,
             "SIDE_COL": 9,
+            "PARAMS_COL": 10,
         }
 
         self.bom_pos_icons = [
@@ -74,6 +77,7 @@ class PartListDataModel(dv.PyDataViewModel):
             "wxDataViewIconText",
             "string",
             "wxDataViewIconText",
+            "string",
         )
         return columntypes[col]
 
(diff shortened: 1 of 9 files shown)


Bouni:
I tested your PR, and I like it so far.
There were some ruff formatter issues that I fixed.

The double click event on the part selector is not working (at least on Windows 11). 
That seems to be impossible https://stackoverflow.com/a/50489083

It works on the mainwindows however because there we use a `wx.DataViewCtrl`, on the part selector we still use a `wx.dataview.DataViewListCtrl`.
Maybe I'll rewrite that so that we use a `wx.DataViewCtrl` as well and get a double click event.

But at te moment I hav no time for that 😓

Bouni:
I'll merge this as it worked perfectly fine for me, the only thing that needs some work is the doubleclick feature in the partselector.
I will open a new issue for that.

wrs:
@Bouni I didn't have notifications turned on for some reason — thanks for the merge!

I just noticed that sorting in the table view isn't working…not sure if this broke it or not. When I get a chance I'll figure it out and send a fix.

PR #555, opened 2024-11-26
Fix: check_order_number function on empty files

andy9a9:
When empty board is used, `pcbnew.GetBoard().GetFileName()` returns `FileNotFound` exception which was not handled in `check_order_number` function. Catch `OSError` exception which should fix this behaviour.

Maybe longing an error would be useful?


Bouni:
Good catch. Thanks for the PR!

Issue #566, opened 2025-01-22
LCSC values not removed from board


TAMPCO:
Hi everyone ! Not sure if that's just me or if the issue is common to everyone, so here it goes.

Problem : after setting an LCSC value to a component, a custom LCSC field appears for that specific component at board level. If you remove LCSC number from the plugin's interface, then close the interface and open it up again, the board's LCSC value has not been removed and now appears again. This behaviour prevents

1. Create a new project, add a component to the schematic and update the board to place the component.
2. In the PCB editor, open Bouni's plugin and assign an LCSC component to the database.
3. Close the plugin interface, then open it again and click on "remove LCSC number" for the component.
4. Close the plugin interface and open it again : the LCSC number has now been taken from the board.

Expected behaviour :
When removing the LCSC value on the plugin interface, value should also be removed from the custom field of the component in the board.

Side note : when removing the LCSC value, the LCSC number, type and stock numbers will not be removed immediately, only if you click somewhere else. Also, the new LCSC Params column is not removed at all, even if you click somewhere else. Should I open another issue for this visual problems ? I'm not really used to using Github a lot :)

Thank you very much for your help !
Best regards

```
Application: KiCad x64 on x64

Version: 8.0.8, release build

Libraries:
	wxWidgets 3.2.6
	FreeType 2.13.3
	HarfBuzz 10.0.1
	FontConfig 2.15.0
	libcurl/8.10.1-DEV Schannel zlib/1.3.1

Platform: Windows 11 (version 26100), édition 64-bit, 64 bit, Little endian, wxMSW

Build Info:
	Date: Jan 11 2025 18:59:49
	wxWidgets: 3.2.6 (wchar_t,wx containers)
	Boost: 1.86.0
	OCC: 7.8.1
	Curl: 8.10.1-DEV
	ngspice: 44
	Compiler: Visual C++ 1942 without C++ ABI

Build settings:

```

Bouni:
Would you mind submiting a PR? 
I will test it as soon as my free time allows me to do so and merge it if everything works fine.

z2amiller:
Fixed by: #569

PR #569, opened 2025-02-04
Fixed params display in mainwindow and LCSC board value when removed

TAMPCO:
See issue #566. With this commit :
- When LCSC value is removed from a component in the main window, the params display gets updated as well.
- When LCSC value is removed from a component in the main window, the footprint's LCSC field is emptied as well, preventing the footprint from reloading its previous LCSC value into the database.


Bouni:
Thanks for the PR and sorry for the late reply an merge

PR #577, opened 2025-03-05
PartSelector refactoring

Bouni:
Use a DataViewCtrl instead of a DataViewListCtrl.

- better sorting
- Double Click selection of a part
- Datamodel in the background

These changes are approximately the smae as already done in the mainwindow a while ago


chmorgan (approved):
Tested here, the dialogs look much sharper. Tested sorting, part search etc.

Bouni:
Thanks for testing, highly appreciated 👍

PR #580, opened 2025-03-11
refactor schematic version handling

### LINKED ISSUE #579: "Export to schematic" doesn't work in KiCad 9
**Describe the bug**
Exporting to schematic appears to work, but fails to update the LCSC field

**To Reproduce**
Press "Export to Schematic" after editing values

**Expected behavior**
Values should update

# KiCad Version
```
Application: KiCad arm64 on arm64
Version: 9.0.0, release build
Libraries:
    wxWidgets 3.2.6
    FreeType 2.13.3
    HarfBuzz 10.2.0
    FontConfig 2.15.0
    libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.62.0
Platform: macOS Sequoia Version 15.1.1 (Build 24B91), 64 bit, Little endian, wxMac
OpenGL: Apple, Apple M1 Pro, 2.1 Metal - 89.3
Build Info:
    Date: Feb 19 2025 13:05:44
    wxWidgets: 3.2.6 (wchar_t,wx containers)
    Boost: 1.87.0
    OCC: 7.8.1
    Curl: 8.7.1
    ngspice: 44.2
    Compiler: Clang 16.0.0 with C++ ABI 1002
    KICAD_IPC_API=ON
```

andy9a9:
- Invert the version logic. Try to search for versions 6 and 7 explicitly. For versions 8+ use the default mechanism. This will allow us to be compatible with newly builds.
- Disable "is_version8/9" methods, but keep them ready when needed

Closes #579

### DIFF
--- a/helpers.py
+++ b/helpers.py
@@ -1,5 +1,6 @@
 """Contains helper function used all over the plugin."""
 
+from distutils.version import LooseVersion
 import os
 import re
 
@@ -12,19 +13,30 @@ EXCLUDE_FROM_POS = 2
 EXCLUDE_FROM_BOM = 3
 
 
-def is_version8(version: str) -> bool:
-    """Check if version is 8 or 8 Nightly build."""
-    return any(v in version for v in ("8.0", "8.99"))
+def _is_version_in_range(version: str, min_version: str, max_version: str) -> bool:
+    """Check if version is in range."""
+    ver = LooseVersion(version)
+    return LooseVersion(min_version) <= ver < LooseVersion(max_version)
+
+
+# def is_version9(version: str) -> bool:
+#    """Check if version is 9."""
+#    return _is_version_in_range(version, "8.99", "10.0")
+
+
+# def is_version8(version: str) -> bool:
+#    """Check if version is 8."""
+#    return _is_version_in_range(version, "7.99", "9.0")
 
 
 def is_version7(version: str) -> bool:
     """Check if version is 7."""
-    return any(v in version for v in ("7.0"))
+    return _is_version_in_range(version, "6.99", "8.0")
 
 
 def is_version6(version: str) -> bool:
     """Check if version is 6."""
-    return any(v in version for v in ("6.0"))
+    return _is_version_in_range(version, "5.99", "7.0")
 
 
 def getWxWidgetsVersion():
@@ -146,6 +158,7 @@ def set_lcsc_value(fp, lcsc: str):
         fp.SetField("LCSC", lcsc)
         field = fp.GetFieldByName("LCSC")
         field.SetVisible(False)
+
 
 def get_valid_footprints(board):
     """Get all footprints that have a valid reference (drop all REF**)."""
--- a/schematicexport.py
+++ b/schematicexport.py
@@ -7,7 +7,7 @@ import re
 
 from pcbnew import GetBuildVersion  # pylint: disable=import-error
 
-from .helpers import is_version6, is_version7, is_version8
+from .helpers import is_version6, is_version7
 
 
 class SchematicExport:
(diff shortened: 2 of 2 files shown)


nickoe:
@andy9a9  Cool that you tried to fix this, but this does not work for me. When I cherry-pick your change the plugin does not load. I don't see any error on load, possibly because this is on windows. Although I did see something printed from another plugin.

Version is printed as: `Version: 9.0.0, release build` and in python console:
```
pcbnew.GetBaseVersion()
'9.0.0'
```

andy9a9:
Hi @nickoe ,
> Version is printed as: `Version: 9.0.0, release build` and in python console:
> 
> ```
> pcbnew.GetBaseVersion()
> '9.0.0'
> pcbnew.GetMajorMinorVersion()
> '9.0'
> ```

Where did you find this code? Is it a part of your plugin?
What does it mean, not load?
Could you please open the `wxPython` console before trying to load the plugin?

nickoe:
I just openend the python console in pcbnew and import pcbnew and the other commands

andy9a9:
> I just openend the python console in pcbnew and import pcbnew and the other commands

Not sure if I get it correctly, but how did you use this plugin?
Did you open the python console in `pcbnew` and clicked on the jlc-plugin icon?

nickoe:
@andy9a9  I am not sure what you are concerned about in this case. I just open it via the action in pcbnew. Tools -> External Plugins > JLCPCB Tools.  

I started at described in the README to debug the plugin and test your patch.

Your patch would not work on the windows build and possibly also fail on some distributions on linux. For the MSVC build the packaging module is not included, hence it fails to load because it can't find it on early load of the plugin.  And I am not sure if you can rely on that non-standard module to be available generally. It is risky to use that this early without a try expect that logs something.

I made an alternative path like so that works on windows:
```
diff --git a/helpers.py b/helpers.py
index d9941da..a5e9f06 100644
--- a/helpers.py
+++ b/helpers.py
@@ -12,9 +12,9 @@ EXCLUDE_FROM_POS = 2
 EXCLUDE_FROM_BOM = 3
 
 
-def is_version8(version: str) -> bool:
+def is_version8_or_version9(version: str) -> bool:
     """Check if version is 8 or 8 Nightly build."""
-    return any(v in version for v in ("8.0", "8.99"))
+    return any(v in version for v in ("8.0", "8.99", "9.0", "9.99"))
 
 
 def is_version7(version: str) -> bool:
diff --git a/schematicexport.py b/schematicexport.py
index 40343d0..bc7ba4f 100644
--- a/schematicexport.py
+++ b/schematicexport.py
@@ -5,9 +5,9 @@ import os
 import os.path
 import re
 
-from pcbnew import GetBuildVersion  # pylint: disable=import-error
+from pcbnew import GetBaseVersion  # pylint: disable=import-error
 
-from .helpers import is_version6, is_version7, is_version8
+from .helpers import is_version6, is_version7, is_version8_or_version9
 
 
 class SchematicExport:
@@ -21,18 +21,20 @@ class SchematicExport:
 
     def load_schematic(self, paths):
         """Load schematic file."""
-        if is_version8(GetBuildVersion()):
-            self.logger.info("Kicad 8+...")
+        if is_version8_or_version9(GetBaseVersion()):
+            self.logger.info("Kicad 8 or 9...")
             for path in paths:
                 self._update_schematic8(path)
-        elif is_version7(GetBuildVersion()):
+        elif is_version7(GetBaseVersion()):
             self.logger.info("Kicad 7...")
             for path in paths:
                 self._update_schematic7(path)
-        elif is_version6(GetBuildVersion()):
+        elif is_version6(GetBaseVersion()):
             self.logger.info("Kicad 6...")
             for path in paths:
                 self._update_schematic(path)
+        else:
+            self.logger.error("Did not detect a supported Kicad version. Please update me!")
 
     def _update_schematic(self, path):
         """Only works with KiCad V6 files."""
-- 
2.43.0
```

andy9a9:
@nickoe , I think your variant is step back to mention all combinations. But anyway thanks for the hint, that it doesn't work on the windows, so I refactored it.

schematicexport.py
```diff
@@ -185,7 +185,7 @@ def _update_schematic7(self, path):
                 f.write(line + "\n")
         self.logger.info("Added LCSC's to %s (maybe?)", path)
 
-    def _update_schematic8(self, path):
+    def _update_schematic(self, path):
```
chmorgan, line 188: can we make this _update_schematic8 for consistency and because we've got kicad 9 here and likely want to do v9 specific things.

helpers.py
```diff
@@ -12,19 +13,30 @@
...
-def is_version8(version: str) -> bool:
-    """Check if version is 8 or 8 Nightly build."""
-    return any(v in version for v in ("8.0", "8.99"))
+def _is_version_in_range(version: str, min_version: str, max_version: str) -> bool:
+    """Check if version is in range."""
+    ver = LooseVersion(version)
+    return LooseVersion(min_version) <= ver < LooseVersion(max_version)
+
+
+# def is_version9(version: str) -> bool:
```
chmorgan, line 22: if these aren't being used its ok to remove them, I think is_version6 and is_version7 are suitable templates for a v8 or v9. version.

chmorgan (changes requested):

schematicexport.py
andy9a9: I wanted to invert the logic and keep the variant compatible with latest versions. When the versions 8/9 won't share the code anymore, we can split it into separated functions.

helpers.py
andy9a9: I guess this future preparation won't harm.. If we wanted to distinguish between versions we don't have to search for the specifics numbers.

chmorgan (approved):

chmorgan:
Merged with some minor edits.

Issue #525, opened 2024-08-15
Feature Request: Export Additional Layers


lilindian16:
I am in the process of developing flex PCBs. JLC requires a dedicated layer for outlining the Stainless Steel stiffener with the specs included in the zipped gerber. My current work-around is to draw the details on the User.Comments layer and then edit _kicad-jlcpcb-tools/fabircation.py_ by adding ("Stiffener", Cmts_User, "Stainless Steel Stiffener") to the plot_plan_top list in the _generate_geber(self, layer_count=None)_ function. This exports the User.Comments layer as the Stiffener layer.

It would be good to have the option to include additional layers for fabrication file generation in this toolchain. That way we can keep a dedicated and named layer in the editor that we can also easily export - something like a tick-box with the default exported layers already selected. Kicad includes this feature in the built-in gerber exporter

 ![image](https://github.com/user-attachments/assets/85239e4e-6bc1-42a4-b996-4722347e7716)

Bouni:
I see your point but strugle to see how we could easily include that so that the normal user does not get affected.
I don't want to replicate the exporter settings of KiCAD as the whole point of the plugin is that you can press export and get what JLC needs.

Maybe we could add a setting "export User layers" which then exports all user layers into the gebrer zip.
For sure there are to many in there for your use case but you could still delete the not need ones from the zip.

Do you have a better idea than that?

zevv:
A viable alternative would be to have a whitelist of 'special' layers that will be exported if these exist, or export all layers that have a specific prefix or suffix (for example `JLC_`): one could then name a layer "JLC_stiffener_BOT_0.2_mm_PI" to have it exported.

Bouni:
I will think about a way to realize this. Renaming layers seems a bit inconvenient. 
Maybe a new Setting that lets the user choos which layers should be exported and a button to reset to default!?

zevv:
> I will think about a way to realize this. Renaming layers seems a bit inconvenient. Maybe a new Setting that lets the user choos which layers should be exported and a button to reset to default!?

Will these settings be persistent so they will be stored as part of the project? It would be nice if the right layers are still exported if a colleague orders a new batch next year :)

lilindian16:
[...]

PR #581, opened 2025-03-12
Feature Request: Export Additional Layers #525

lilindian16:
Added support for automatically exporting layers with _JLC__ prefix in the layer name. User must add a user defined layer with _JLC__ as the name prefix e.g. _JLC_SS_Stiffener_Top_0.2mm_

As of now there would be no way to disable this - if you have a layer with the _JLC__ prefix it will always be exported in your gerber. I believe this feature would only be used by people intentionally creating the JLC_ layer so there is probably no need to enable/disable it with a setting? 

Note: the _JLC__ prefix is not case sensitive because I know I would eventually write something like JlC_ or JLc_ by accident. 

Let me know what you think :)

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -25,6 +25,7 @@ from pcbnew import (  # pylint: disable=import-error
     F_SilkS,
     Refresh,
     ToMM,
+    ToLAYER_ID
 )
 
 # Compatibility hack for V6 / V7 / V7.99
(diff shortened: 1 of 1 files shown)


chmorgan:
The last two commits are confusing, you remove an entry, lots of formatting is updated, then a commit to undo that formatting. Possible to fold those two commits together with a rebase? If you aren't sure how I can do so before this is merged.

PR #583, opened 2025-03-23
Extended rotation correction to also have X/Y offsets.

### LINKED ISSUE #418: Placement offsets for parts
**Is your feature request related to a problem? Please describe.**

Some footprints I use result in the parts being placed rotated and with a offset in the JLCPCB placement preview.
For rotation, there is a setting in the plugin but for position offsets, there isn’t (as far as I can tell).

**Describe the solution you'd like**

I’d like to have the option to store a offset that should be used when adding a part to the POS file. Just like the rotation.
Specifying an offset as a relative value (move by dx/dy) would be fine.
Of course there could be other options that simplify the entry:

* Place based on center of footprint (ignoring what layers?)
* Place based on center of pin 1
* Place based on center of all pins
* Place based on top/bottom left/right corner
* …

**Describe alternatives you've considered**

* _Changing the footprints_
  Surely an option for custom footprints but not ideal for standard-parts (e.g. Connector_USB:USB_C_Receptacle_G-Switch_GT-USB-7010ASV vs. C963370)
* _Moving the parts in the placement preview when ordering_
   Works but can easily be forgotten. Especially for parts like the mentioned USB-C connector where the offset is small.

**Additional context**

-/-

dhoepfl:
Implements/Fixes #418

Renamed the rotations.db file to corrections.db to reflect the more general correction approach.
Implemented update logic to convert existing rotation databases to the new correction database (setting the offsets to 0 in this case).
The online corrections file already contains offsets, these are imported.

Some notes about things I am unsure about:

* While implementing this, I noticed that there is a `rotation` table created in the parts.db. This table seems to be never used thus I removed the table creation. (see change of jlcparts_db_convert.py)
* I’m not 100% sure if the code works with KiCAD before version 8, I especially unsure if the usage of wxPoint in fabrication.py is the correct way to do what I do there.
* The new code automatically imports the previous rotations.db. I tried to modify all code (e.g.  import/export) in a way that does not break existing workflows. Obviously exporting now includes two additional columns but importing still accepts files that do not contain offset information (setting those to 0/0).
* I tried to match the code style as best as I could but Python not my day-to-day language so please bear with me.

### DIFF
--- a/corrections.py
+++ b/corrections.py
@@ -1,4 +1,4 @@
-"""Contains the rotations manager."""
+"""Contains the corrections manager."""
 
 import csv
 import logging
@@ -12,15 +12,15 @@ from .events import PopulateFootprintListEvent
 from .helpers import PLUGIN_PATH, HighResWxSize, loadBitmapScaled
 
 
-class RotationManagerDialog(wx.Dialog):
-    """Dialog for managing part rotations."""
+class CorrectionManagerDialog(wx.Dialog):
+    """Dialog for managing part corrections."""
 
     def __init__(self, parent, footprint):
         wx.Dialog.__init__(
             self,
             parent,
             id=wx.ID_ANY,
-            title="Rotations Manager",
+            title="Corrections Manager",
             pos=wx.DefaultPosition,
             size=HighResWxSize(parent.window, wx.Size(800, 800)),
             style=wx.DEFAULT_DIALOG_STYLE | wx.RESIZE_BORDER | wx.MAXIMIZE_BOX,
@@ -29,7 +29,9 @@ class RotationManagerDialog(wx.Dialog):
         self.logger = logging.getLogger(__name__)
         self.parent = parent
         self.selection_regex = None
-        self.selection_correction = None
+        self.selection_rotation = None
+        self.selection_offset_x = None
+        self.selection_offset_y = None
         self.import_legacy_corrections()
 
         # ---------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


chmorgan:
"While implementing this, I noticed that there is a rotation table created in the parts.db. This table seems to be never used thus I removed the table creation. (see change of jlcparts_db_convert.py)"

This makes sense. Can you open a PR with just this individual change so it can be discussed in isolation?

chmorgan:
@dhoepfl can you explain how one would arrive at the desired offsets, eg. what is the workflow to be used? That would be a helpful addition to the README even if only text without images.

dhoepfl:
> This makes sense. Can you open a PR with just this individual change so it can be discussed in isolation?

Sure. See #586

dhoepfl:
> @dhoepfl can you explain how one would arrive at the desired offsets, eg. what is the workflow to be used? That would be a helpful addition to the README even if only text without images.

The [list used for the download button](https://raw.githubusercontent.com/matthewlai/JLCKicadTools/master/jlc_kicad_tools/cpl_rotations_db.csv) already contains correction offsets (e.g. for “USB_C_Receptacle_XKB_U262-16XN-4BVC11”) so if you use that, you will automatically get them.

In my current project, I use one custom footprint that made me write the offset correction: I have no idea what goes wrong in get_position, the pads in the footprint are positioned around the origin, different sizes but the outer edges of the pads are all +/-1.2 mm from the origin. Nonetheless, creating the CPL moves it to the left by 0.18mm. Would probably have worked but I didn’t want to take a risk.

(Something similar happens with the standard “Texas_RWU0007A_VQFN-7_2x2mm_P0.5mm” footprint: Place it at 100/100 and the CPL contains 99.99975/-100.0, might be rounding in this case but I have seen bigger differences.)

So, long story short, I do not have a great solution other than creating the production files, looking closely at them and after uploading them to JLCPCB, looking at what they show in their preview (especially the coordinates at the bottom right are helpful).

Honestly, I am no longer sure if the offset approach is the correct thing to do in all cases: My custom footprint is at 98.79595mm (don’t ask, I just ended up there), it gets exported with that value if I use the 0.18 offset. Moving it to X=100 exports it at 100.18 (offset backfiring …). Moving the chip to a different location just to get the correct coordinates is not an option, so using a project local correction file works for me.

chmorgan:
@dhoepfl are the offsets consistent between two boards with the same part? (that would suggest a global offset table).

I went through the CPL generation path and other than calling ToMM() the only other function that could be causing an issue is this possible aux_origin:

```
    def generate_cpl(self):
        """Generate placement file (CPL)."""
        cplname = f"CPL-{Path(self.filename).stem}.csv"
        self.corrections = self.parent.library.get_all_correction_data()
        aux_orgin = self.board.GetDesignSettings().GetAuxOrigin()
        add_without_lcsc = self.parent.settings.get("gerber", {}).get(
            "lcsc_bom_cpl", True
        )
        with open(
            os.path.join(self.outputdir, cplname), "w", newline="", encoding="utf-8"
        ) as csvfile:
            writer = csv.writer(csvfile, delimiter=",")
            writer.writerow(
                ["Designator", "Val", "Package", "Mid X", "Mid Y", "Rotation", "Layer"]
            )
            footprints = sorted(self.board.Footprints(), key=lambda x: x.GetReference())
            for fp in footprints:
                part = self.parent.store.get_part(fp.GetReference())
                if not part:  # No matching part in the database, continue
                    continue
                if part["exclude_from_pos"] == 1:
                    continue
                if not add_without_lcsc and not part["lcsc"]:
                    continue
                try:  # Kicad <= 8.0
                    position = self.get_position(fp) - aux_orgin
                except TypeError:  # Kicad 8.99
                    x1, y1 = self.get_position(fp)
                    x2, y2 = aux_orgin
                    position = VECTOR2I(x1 - x2, y1 - y2)
                writer.writerow(
                    [
                        part["reference"],
                        part["value"],
                        part["footprint"],
                        ToMM(position.x),
                        ToMM(position.y) * -1,
                        self.fix_rotation(fp),
                        "top" if fp.GetLayer() == 0 else "bottom",
                    ]
                )
        self.logger.info(
            "Finished generating CPL file %s", os.path.join(self.outputdir, cplname)
        )
```

The 0.18mm offset could be a separate thing than part offsets.

Sometimes parts are centered on pin 1, in other cases it's the centroid (the middle of the package). Could that represent the reason why jlcpcb gets the positions incorrect? (I've totally seen the same thing and they just fix them for me on each push, perhaps manually?)

Could you make a test project and push it up to GitHub that has parts from the library and exhibits the offset that you see, using metric units (to ensure there isn't some rounding issue), and with parts on whole mm numbers?

We want to rule out the plugin messing something up. It seems unlikely that the issue is in kicad itself, the coordinates are used all over in pcbnew.

dhoepfl:
> Could you make a test project and push it up to GitHub that has parts from the library and exhibits the offset that you see, using metric units (to ensure there isn't some rounding issue), and with parts on whole mm numbers?

While trying to do so, I found the problem with “Texas_RWU0007A_VQFN-7_2x2mm_P0.5m”: There are 6 (invisible) pads on the pads 1-3. The ones on the left are a tiny bit larger than the pads 1-3 so the calculated center is of by half of this (0.00025). Using the measure tool does not show this because it only shows 1/1000th of a mm.

So probably the export is all fine, I probably confused the 0.18 offset that JLCPCB shows in their preview with the (now explained) 0.00025 offset I saw in the CPL vs. the position I used for the part. 

I faintly remember that the reason I used my own footprint vs. the standard one were these additional pads that I do not need and the offset they caused. But I couldn’t understand why I still see an offset when compared to the JLCPCB preview. Probably I am mixing things up here.

I am very sorry for all the confusion I cause.

To clean the mess up, here’s my current understanding:

* There is no bug in the export. The 0.00025 offset is to be explained by the additional pads.
* The 0.18 offset I see is to be explained by JLCPCB using a different center point. (See “JLCPCB Preview.png” in the [attached file](https://github.com/user-attachments/files/19484873/Offset.Test.zip)), having the option to add a correction makes sense in that case.

dhoepfl:
> so if they used the fab they should be centered on the origin, the fab is 1mm and 1mm across the X.
> 
> If they used the pads extents they should be centered on the origin, they are 1.2mm each side.
> 
> If they used the extents of the silk screen you'd expect more bias.
> 
> Thoughts on if they are using something in the design package relative to the part to determine the center point?
> 
> And, since we have this test design package, could you ask them how they do it? They might tell us its some internal thing and then we've confirmed there should be offsets, or they might say it something in the design package and we can automatically fix it.

I sent a mail to JLCPCB asking for how they do it. Here’s the answer:

> Thanks for your email!
>
>The component's placement page is for reference purpose, which is not 100% perfect. And our engineer actually prepare parts placement based on silkscreen, instead of original parts placement.
>
> Given above circumstance as above, it is suggested to choose "confirm parts placement" button to confirm our parts placement before production.
>
> Our parts placement data is adjustable before production, don't worry.

My conclusions are:

* The silk screen is way more important that I thought it is. (And there is probably no way to find the “correct” location programmatically.)
* As long as the position is within the bounds of the part, they will probably place it correctly.

Regarding the “don’t worry”: The most recent order at work had totally messed up rotation data (newly installed KiCAD, never downloaded the correction database, so no correction was used at all). JLCPCB asked for confirmation of the LED polarity. Just about the *only* parts that were correctly rotated (but not marked on the silk screen) …

dhoepfl:
> That’s a ton of useful info.
> 
> @dhoepfl given this, what do you think the plugin should do? How should we proceed with your pull request?

Obviously I would like to get it. Even if it is not required to get the previews to line up, it would simplify things for me (in my work context, where the orders are placed by the purchasing department).

The arguments I see that are against merging it:

* It seems like it is not really required at the moment. (My point: While it might not be required right now, there might be cases where it helps (see above). Having a 100% correct preview does not break anything, after all.)
* From your point of view: The burden of maintenance.
* While there are corrections in the downloaded default list, I just tested the USB-C plug (using C22383468) which has a x-offset of 1.44mm. It seems like the offset should be 0 now. I assume that wasn’t the case in the past.

So, I see there are reasons not to merge it but if you ask me: Please merge it.

chmorgan:
@dhoepfl I agree with your assessment, I figured you'd articulate it better than I could guess about it :-)

Would you mind one final question for jlcpcb? I know it could delay this slightly but I'm hoping that's ok. The question is, "would they be ok if this plugin adjusted the positions to center them in the pcb preview?" It would stink if they did start using the part center for something or if they already do use it for something else.

dhoepfl:
> @dhoepfl I agree with your assessment, I figured you'd articulate it better than I could guess about it :-)
> 
> 
> 
> Would you mind one final question for jlcpcb? I know it could delay this slightly but I'm hoping that's ok. The question is, "would they be ok if this plugin adjusted the positions to center them in the pcb preview?" It would stink if they did start using the part center for something or if they already do use it for something else.

No worries, the last days were kind of busy anyways, I didn't even find time to return to JLCPCB's mail to send the follow up question. I hope I'll find some time this weekend to do so.

dhoepfl:
Here’s JLCPCB’s answer:

>> Thanks a lot for these information. There is one follow up
   question from the plugin maintainer: "would they be ok if this plugin
   adjusted the positions to center them in the pcb preview?" Thanks a lot
   for your support!
>
>    The component's placement page is for reference purpose , so I think the
   plugin parts placement issue you mention doesn't matter.
>
>   Our engineer will correct placement and prepare DFM based on silkscreen.
>
>   It is suggested to choose "confirm parts placement button" for your concern.

My interpretation is: As long as the part is at “about the right position” it will work.

chmorgan:
Hi @dhoepfl rebased and merged into main. Thank you for confirming that adjustments aren't going to make life difficult for jlcpcb.

PR #584, opened 2025-03-23
Board local corrections

dhoepfl:
Based on my other pull request (#583), I added an option to use project local corrections: By default the global corrections database is used but you can choose to use project local corrections. When doing so, the then current global corrections are copied into the project.db database and that data is used.

Returning to the global corrections database simply drops the correction table from the project.db (with an appropriate warning since all local corrections are lost).

The logic whether the local or global correction data is used is based on the existence of the correction table in project.db. If it exists, even empty, local correction data is used, if it does not exist, global correction data is used.

If you want to transfer the local corrections into the global corrections database, you can use export/import to do so.

Alternative considered: In my first attempt, I used a dedicated corrections.db file next to the project.db. I didn’t see any advantages in doing so.

### DIFF
--- a/corrections.py
+++ b/corrections.py
@@ -441,7 +441,7 @@ class CorrectionManagerDialog(wx.Dialog):
                 wx.YES_NO | wx.NO_DEFAULT | wx.ICON_WARNING
             )
         result = dialog.ShowModal()
-        
+
         if result == wx.ID_NO:
             self.global_corrections.SetValue(self.parent.library.uses_global_correction_database())
             return
--- a/corrections.py
+++ b/corrections.py
@@ -293,12 +293,31 @@ class CorrectionManagerDialog(wx.Dialog):
         )
         self.export_button.SetBitmapMargins((2, 0))
 
+        self.global_corrections = wx.CheckBox(
+            self,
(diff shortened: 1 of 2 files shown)


chmorgan:
@Bouni thoughts on including these inside of project.db? I'd even propose folding the global mappings.db and rotations.db into a global.db to avoid proliferating separate database files since we've got the concept of tables inside of a single database...

chmorgan:
@dhoepfl why would someone want board local rotation/position corrections vs. the global ones?

dhoepfl:
> @dhoepfl why would someone want board local rotation/position corrections vs. the global ones?

The rare reason (that bite me when I updated from 8.x to 9.0): The global databases are lost when you uninstall, then reinstall the plugin or when you update KiCAD, then delete the old plugins directory without moving them over.

The main reason: You can check in the local corrections database and others that clone the repository get your corrections (e.g. for project local footprints). Same if you send the PCB design files to a third party.

Bouni:
LGTM

In my opinion we should think of a way to persist the corrections so that a update does not purge them ...

And maybe write the matching ones to the project anyway!?

PR #586, opened 2025-03-26
Removed unused 'rotation' table from the parts database.

dhoepfl:
The 'rotation' table in the parts database is created but I cannot find any usage of it.
So I removed it.


chmorgan:
Thank you for your contribution!

PR #588, opened 2025-03-26
Feature/jlc specific layers

lilindian16:
- Added support for automatically exporting layers with JLC_ prefix in the layer name. User must add a user defined layer with JLC_ as the name prefix e.g. JLC_SS_Stiffener_Top_0.2mm

- Added instructions for using the feature in readme

As of now there would be no way to disable this - if you have a layer with the JLC_ prefix it will always be exported in your gerber. I believe this feature would only be used by people intentionally creating the JLC_ layer so there is probably no need to enable/disable it with a setting?

Note: the JLC_ prefix is not case sensitive because I know I would eventually write something like JlC_ or JLc_ by accident.

Let me know what you think :)


chmorgan:
Thank you for your contribution!

PR #603, opened 2025-04-07
Cmm db optimize

chmorgan:


Bouni:
Can you give a bit context what this PR does / improves?

Does it primarily improve the generation of the database?

chmorgan:
> Can you give a bit context what this PR does / improves?
> 
> Does it primarily improve the generation of the database?

- ~25% reduction in database size (4.2GB to 3.1GB) (100MB reduction in compressed size 592MB to 474MB)
   This should speed up search time although I haven't benchmarked it yet.
- Introduced pytests for the jlcparts_db_convert.py in CI. We should be able to more easily add tests for other modules as they are built out.
- Moved the parts conversion into db_build/ so it can be a module.

chmorgan:
@Bouni and there are details in the commit messages for each commit, relative to the db changes and in code and in the commits relative to the price changes.

Bouni (approved):

PR #608, opened 2025-04-11
Added some input sanitization to prevent database corruption.

dhoepfl:
I noticed that e.g. setting a rotation of '90x' resulted in a unreadable correction table.
(No issue to link this pull request to because I already have the fix …)

Please review the code, I am not sure if it matches your code style.
I tried my best but python still is not my main language.


Bouni:
Can you add the two docstrings please?

Issue #602, opened 2025-04-03
Different Rotation displayed on the window vs in the exported CPL file


debreczeni:
**Describe the bug**
In the case described below the UI window displays a different value for the rotation compared to what it is exporting. 

My original issue was that the rotation regexes didn't work because of their order dependency until I noticed that the CSV export seems to be OK when overriding by component name, but the main UI window seems to be displaying a different value so the regexes are matched in a different order or with a different logic.

**To Reproduce**
1. Setup an override for a rotation by name that would otherwise get rotated by a footprint match (a particular naming or regex order might be required to reproduce this - In my case the footprint match was the rule for `SOT-23` matching `SOT-23-TH-16`, and overridden by name rule for `TL074HIDYYR` as overriding with the footprint didn't do anything.
2. Click on Save and Update in the rotations window
3. Exit rotations window
4. Generate export, and cross-check the different rotation values on the main window and the generated CSV

**Expected behavior**
Rotations should match normally - they are different in my case as can be seen on the screenshots 

Being able to define a regex matching order could be a plus - but this should get its own issue.

<img width="912" alt="Image" src="https://github.com/user-attachments/assets/c7c3ee03-fe9d-41ab-a6d7-c91e4d8b18da" />

<img width="1180" alt="Image" src="https://github.com/user-attachments/assets/9c1eed9a-7abb-499a-afd1-62e860c08bae" />

<img width="699" alt="Image" src="https://github.com/user-attachments/assets/82b547ba-ca78-4bef-916f-960f5349540d" />

# JLCPCB Toolkit Version
Most recent version, as of today: `2025-04-02`
It had the same behaviour on  version `2024-11-01` too.

# KiCad Version

```
Application: KiCad PCB Editor arm64 on arm64
Version: 8.0.9, release build
Libraries:
    wxWidgets 3.2.6
    FreeType 2.13.3
    HarfBuzz 10.2.0
    FontConfig 2.15.0
    libcurl/8.4.0 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.51.0
Platform: macOS Ventura Version 13.6.4 (Build 22G513), 64 bit, Little endian, wxMac
OpenGL: Apple, Apple M1 Max, 2.1 Metal - 83.1
Build Info:
    Date: Feb 18 2025 16:12:55
    wxWidgets: 3.2.6 (wchar_t,wx containers)
    Boost: 1.87.0
    OCC: 7.8.1
[...]

PR #615, opened 2025-04-20
Unified correction matching between main window and fabrication.

dhoepfl:
Fix for issue #602:

* Unified correction matching between main window and fabrication, new order is reference, value, footprint.
* Added an indicator (“(ref)”, ”(val)”, “(fpt)“) to the correction column to show what the correction matched on.

Also includes pull #477 (Added rotation by reference):

* Added context menu item to create correction rule based on reference.

Additionally:

* Changed column title “Value” to “Value (Name)” since value is sometimes called name in the UI elsewhere.


Bouni:
Again, sorry for the months long delay ....

Thanks for your efforts!

PR #625, opened 2025-06-10
Add DNP parameter support for component exclusion

Hizuki1030:
## Summary
- Add support for DNP (Do Not Populate) parameter to exclude components from fabrication

## Test plan
- Test that components with DNP parameter are properly excluded
- Verify existing functionality remains intact

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/helpers.py
+++ b/helpers.py
@@ -134,6 +134,32 @@ def get_lcsc_value(fp):
     return ""
 
 
+def get_dnp_value(fp):
+    """Get DNP (Do Not Populate) parameter from footprint fields or properties."""
+    # Check ${DNP} parameter in KiCad 7.99+
+    try:
+        for field in fp.GetFields():
+            if field.GetName().upper() == "DNP":
+                text = field.GetText().strip()
+                return text.lower() in ("true", "1", "yes", "on")
+    # KiCad <= V7
+    except AttributeError:
+        for key, value in fp.GetProperties().items():
+            if key.upper() == "DNP":
+                return value.strip().lower() in ("true", "1", "yes", "on")
+    
+    # Fallback to native DNP flag if available
+    if hasattr(fp, "IsDNP") and callable(fp.IsDNP):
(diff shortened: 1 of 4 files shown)


MichaelRColton:
Would this address my issue? https://github.com/Bouni/kicad-jlcpcb-tools/issues/660

store.py
```diff
@@ -171,14 +173,26 @@ def set_lcsc(self, ref: str, lcsc: str):
     def update_from_board(self):
         """Read all footprints from the board and insert them into the database if they do not exist."""
         for fp in get_valid_footprints(self.board):
+            # Check DNP parameter and auto-set exclude flags
+            dnp_detected = get_dnp_value(fp)
+            exclude_from_bom = get_exclude_from_bom(fp) or dnp_detected
+            exclude_from_pos = get_exclude_from_pos(fp) or dnp_detected
+
```
chmorgan, line 180: shouldn't we simplify this to avoid needing to expose a get_dnp_value()? Shouldn't it be enough to call get_exclude_from_bom() or get_exclude_from_pos() and they would know whether the footprint fields or properties indicated DNP internally?

helpers.py
```diff
@@ -134,6 +134,32 @@ def get_lcsc_value(fp):
...
+                return value.strip().lower() in ("true", "1", "yes", "on")
+    
+    # Fallback to native DNP flag if available
+    if hasattr(fp, "IsDNP") and callable(fp.IsDNP):
+        return fp.IsDNP()
+    
+    return False
+
+
+def should_exclude_from_bom_or_pos(fp):
```
chmorgan, line 158: why isn't this function:

```
return get_exclude_from_bom(fp) | get_exclude_from_pos(fp)
```

?

chmorgan (changes requested):
Thank you for the contribution. Had a few questions. Could you also rebase if you plan to continue proposing this PR?

Issue #609, opened 2025-04-11
Duplicate entries in the corrections list


dhoepfl:
**Describe the bug**
Saving a correction entry with an regex that already exists in the list of corrections results in duplicate entries in the list.
Pressing “Delete” deletes all entries.
(The database is correct, only the latest entry is present.)

**To Reproduce**
1. Open the corrections dialog
2. Enter a new pattern, e.g. “TEST”
3. Click “Save”
4. Click “Save” again
5. Click “Delete”

**Expected behavior**
Saving a “new” entry with an regex that already exists should update the entry, even when nothing is selected.

Edge case: Entry “a” is selected, its regex is edited to “b”, which *also* exists. What should saving do? Delete “a” and update “b”?

# KiCad Version
Application: KiCad x86_64 on x86_64

Version: 9.0.0, release build

Libraries:
	wxWidgets 3.2.6
	FreeType 2.13.3
	HarfBuzz 10.2.0
	FontConfig 2.15.0
	libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.64.0

Platform: macOS Sequoia Version 15.4 (Build 24E248), 64 bit, Little endian, wxMac

Build Info:
	Date: Feb 19 2025 14:48:01
	wxWidgets: 3.2.6 (wchar_t,wx containers)
	Boost: 1.87.0
	OCC: 7.8.1
	Curl: 8.7.1
	ngspice: 44.2
	Compiler: Clang 16.0.0 with C++ ABI 1002
	KICAD_IPC_API=ON

Bouni:
I hope you'll submit a PR for this 😏

dhoepfl:
> I hope you'll submit a PR for this 😏

Might take a few days.
What do you think about the edge case I described? (Alert: Do (not) replace?)

z2amiller:
Fixed by: #630

PR #630, opened 2025-06-19
Added handling of duplicate entries on saving a correction.

dhoepfl:
The following rules are used when saving:

* Nothing selected, new regex 
  Add a new entry.
* Nothing selected, regex exists, all values equal
  Select row.
* Nothing selected, regex exists, values differ
  Ask user if entry should be updated.
* The selected regex is equal to the edit field
  Update the entry.
* The selected regex is different from the edited one but new
  Update entry (changing the regex).
* The selected regex is different from the edited one but other row for that regex exists
  Ask the user if that entry should be replaced.
  That means the selected entry is deleted, the existing one is updated.

Please note that I changed the correction list selection mode from single row to multiple rows (but unselect all but one row in the selection handler). I did this because I did not find any other way to allow the user to deselect the rows.
One alternative option would be to add a “(Save as) New” button which ignores the selected row (clearing the input fields if they match the selected row - “New”, otherwise saving them - “Save as New”).
In order to keep this pull request to merely fix issue #609, I postponed this and went with the simpler implementation using multi row selection.

### DIFF
--- a/corrections.py
+++ b/corrections.py
@@ -449,40 +449,38 @@ class CorrectionManagerDialog(wx.Dialog):
                     self.selection_regex = regex
                 else:
                     # The regex exists with different values, ask the user what to do.
-                    existing_correction = "({0}°, {1}/{2})".format(
-                        str(existing_rotation),
-                        self.str_from_float(existing_offset_x),
-                        self.str_from_float(existing_offset_y)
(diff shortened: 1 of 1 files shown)


Bouni:
Nice, thank you for submitting this PR!

PR #631, opened 2025-07-05
Cmm jlcparts fetch - move jlcparts database fetching into the jlcparts_db_convert.py script

chmorgan:
This makes it easier to test jlcparts_db_convert.py as otherwise you end up having to find the GitHub workflow and run those steps manually in a shell. Now the GitHub workflow uses the jlcparts_db_convert.py script.


whmountains:
Wow, are we really having AI do code review now?  I thought it was supposed to be the other way around 😂

PR #633, opened 2025-07-16
fix(version): make version checking more robust

andy9a9:
Versions like `9.0.2-2.fc42` or `7.0.0-rc1-378-ge76fd128c3` were causing an errors, so they have to be normalised before comparison.

/cc @slabua

### DIFF
--- a/core/version.py
+++ b/core/version.py
@@ -1,11 +1,14 @@
 """Contains helper function used all over the plugin."""
 
+import re
+
 from packaging.version import Version
 
 
 def _is_version_in_range(version: str, min_version: str, max_version: str) -> bool:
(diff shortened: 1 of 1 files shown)


chmorgan (changes requested):
Can you fold the normalize into the checking routine? Imo it doesn’t make sense to require normalizing on each version check if the internal routine can perform that normalization.

Also, can you add some tests in a  commit prior to your fix that will result in failures? That will let us check thst your fix fixes the issue by first trying with the new tests and  seeing them fail and then passing with your fix.

andy9a9:
> Can you fold the normalize into the checking routine? Imo it doesn’t make sense to require normalizing on each version check if the internal routine can perform that normalization.
> 
> Also, can you add some tests in a commit prior to your fix that will result in failures? That will let us check thst your fix fixes the issue by first trying with the new tests and seeing them fail and then passing with your fix.

like this?

PR #639, opened 2025-08-10
Fix crash in plugin when datasheet url is missing

fbataill:
Add sanity test to avoid crash when datasheet URL is missing


Bouni:
Thank you for your PR, highly appreciate it!

fbataill:
Your are welcome!

-------- Message d'origine --------

PR #642, opened 2025-08-29
Improve keyword prefill: support PART NUMBER, MPN and LCSC fields

PatriceVigier:
### Summary
This PR improves the default keyword prefill when assigning parts in the JLCPCB Tools plugin.

Instead of always falling back to the `Value` field, the plugin now looks (in order of priority) at:
1. **PART NUMBER** (custom field, often used in company libraries)
2. **MPN** (Manufacturer Part Number, if present)
3. **LCSC** (Cxxxxx code)
4. Fallback: the footprint Value

### Motivation
- Many users (including myself) store the "real" reference in `PART NUMBER` or `MPN` fields rather than the generic `Value`.
- This avoids manual re-typing when searching for parts in the JLCPCB database, saving time and reducing errors.
- The change is backward-compatible: if none of these fields are present, the old behavior (`Value`) is preserved.

### Implementation
- Added `_keyword_from_part_number()` helper to check fields case-insensitively.
- Prefill priority order: PART NUMBER → MPN → LCSC → Value.
- Minor refactor for readability.

### Testing
- Tested on KiCad 9 with projects containing:
  - Only Value field → works as before.
  - PART NUMBER field → correctly used as default search.
  - PART NUMBER containing embedded MPN or LCSC → correctly parsed.
  - LCSC field alone → correctly picked up.

### Notes
This change should be fully backward compatible.  
It mainly benefits workflows where users already maintain `PART NUMBER` or `MPN` fields in their schematics/footprints.

---

🙏 Thank you for this great plugin! This patch makes the search workflow much smoother.

### DIFF
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -13,7 +13,6 @@ import wx  # pylint: disable=import-error
 from wx import adv  # pylint: disable=import-error
 import wx.dataview as dv  # pylint: disable=import-error
 
-from .corrections import CorrectionManagerDialog
 from .datamodel import PartListDataModel
 from .derive_params import params_for_part
 from .events import (
@@ -47,6 +46,7 @@ from .library import Library, LibraryState
 from .partdetails import PartDetailsDialog
 from .partmapper import PartMapperManagerDialog
 from .partselector import PartSelectorDialog
+from .rotations import RotationManagerDialog
 from .schematicexport import SchematicExport
 from .settings import SettingsDialog
 from .store import Store
@@ -56,7 +56,7 @@ logging.getLogger("urllib3").setLevel(logging.WARNING)
 
 ID_GENERATE = 0
 ID_LAYERS = 1
-ID_CORRECTIONS = 2
+ID_ROTATIONS = 2
 ID_MAPPINGS = 3
 ID_DOWNLOAD = 4
 ID_SETTINGS = 5
@@ -73,7 +73,6 @@ ID_SAVE_MAPPINGS = 15
 ID_EXPORT_TO_SCHEMATIC = 16
 ID_CONTEXT_MENU_COPY_LCSC = wx.NewIdRef()
 ID_CONTEXT_MENU_PASTE_LCSC = wx.NewIdRef()
-ID_CONTEXT_MENU_ADD_ROT_BY_REFERENCE = wx.NewIdRef()
 ID_CONTEXT_MENU_ADD_ROT_BY_PACKAGE = wx.NewIdRef()
 ID_CONTEXT_MENU_ADD_ROT_BY_NAME = wx.NewIdRef()
 ID_CONTEXT_MENU_FIND_MAPPING = wx.NewIdRef()
@@ -185,11 +184,11 @@ class JLCPCBTools(wx.Dialog):
 
         self.upper_toolbar.AddStretchableSpace()
 
-        self.correction_button = self.upper_toolbar.AddTool(
-            ID_CORRECTIONS,
-            "Corrections",
+        self.rotation_button = self.upper_toolbar.AddTool(
+            ID_ROTATIONS,
+            "Rotations",
             loadBitmapScaled("mdi-format-rotate-90.png", self.scale_factor),
-            "Manage part corrections",
+            "Manage part rotations",
         )
 
         self.mapping_button = self.upper_toolbar.AddTool(
@@ -218,7 +217,7 @@ class JLCPCBTools(wx.Dialog):
         self.upper_toolbar.Realize()
 
         self.Bind(wx.EVT_TOOL, self.generate_fabrication_data, self.generate_button)
-        self.Bind(wx.EVT_TOOL, self.manage_corrections, self.correction_button)
+        self.Bind(wx.EVT_TOOL, self.manage_rotations, self.rotation_button)
         self.Bind(wx.EVT_TOOL, self.manage_mappings, self.mapping_button)
         self.Bind(wx.EVT_TOOL, self.update_library, self.download_button)
         self.Bind(wx.EVT_TOOL, self.manage_settings, self.settings_button)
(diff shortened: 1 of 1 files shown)


mainwindow.py
```diff
@@ -47,6 +46,7 @@
 from .partdetails import PartDetailsDialog
 from .partmapper import PartMapperManagerDialog
 from .partselector import PartSelectorDialog
+from .rotations import RotationManagerDialog
```
chmorgan, line 49: [P1] Restore the existing correction dialog import

The PR tree contains `corrections.py` and `CorrectionManagerDialog`, but no `rotations.py`. This unconditional import therefore raises `ModuleNotFoundError` on a clean installation. Since `__init__.py` imports `plugin.py`, which imports `mainwindow.py`, the plugin fails before it can register or open. Executing this import against the PR snapshot reproduces the failure. Please retain the existing correction dialog import and its call sites.

mainwindow.py
```diff
@@ -640,18 +628,14 @@ def display_message(self, e):
...
+        for regex, correction in corrections:
             if re.search(regex, str(part["reference"])):
-                return f"{str(rotation)}°, {str(offset[0])}/{str(offset[1])} (ref)"
-        # Then try to match by value
-        for regex, rotation, offset in corrections:
-            if re.search(regex, str(part["value"])):
-                return f"{str(rotation)}°, {str(offset[0])}/{str(offset[1])} (val)"
-        # If there was no match for the part name or value, check if the package matches
-        for regex, rotation, offset in corrections:
+                return str(correction)
```
chmorgan, line 633: [P1] Preserve the three-element correction records

`Library.get_all_correction_data()` still returns `(regex, rotation, offset)` records, so unpacking each record into two variables raises `ValueError: too many values to unpack` whenever any correction is saved. `populate_footprint_list()` calls this method for every displayed footprint, including during startup. With the same correction record and footprint, the parent populates the row while this revision raises before inserting it. This remains broken after the missing import is fixed; please preserve the existing rotation/offset tuple handling and display.

mainwindow.py
```diff
@@ -909,95 +893,112 @@ def select_part(self, *_):
...
-                "There are items with identical LCSC number but different values in the list:\n"
-                + warnings
-                + "Continue?",
-                "Plausibility check",
+                "JLC order number placeholder not present! Continue?",
+                "JLC order number placeholder",
                 wx.OK | wx.CANCEL | wx.CENTER,
             )
-            if result == wx.CANCEL:
+            if result == wx.ID_CANCEL:
```
chmorgan, line 1000: [P2] Compare the MessageBox result with wx.CANCEL

When the placeholder warning appears and the user presses Cancel, `wx.MessageBox()` returns `wx.CANCEL`, whereas this comparison uses `wx.ID_CANCEL`. These are different constants, so the handler continues through zone filling and all fabrication exports despite cancellation. The parent stops in a focused reproduction; this revision calls all six generation steps. Keep the original `wx.CANCEL` comparison here. The distinction from `MessageDialog.ShowModal()` is documented in the [wx.MessageBox API](https://docs.wxpython.org/wx.functions.html#wx.MessageBox).

mainwindow.py
```diff
@@ -909,95 +893,112 @@ def select_part(self, *_):
...
+        return True
 
     def generate_fabrication_data(self, *_):
         """Generate fabrication data."""
-        warnings = self.fabrication.get_part_consistency_warnings()
-        if warnings:
+        if (
+            self.settings.get("general", {}).get("order_number")
+            and not self.check_order_number()
+        ):
```
chmorgan, line 994: [P2] Retain the BOM consistency check before exporting

The previous call to `self.fabrication.get_part_consistency_warnings()` and its confirmation dialog disappear from this handler. For example, with R1 = 10k and R2 = 100k both assigned C25804, the parent warns about the conflicting values and lets the user stop; this revision exports immediately with no warning. The checker still exists in `fabrication.py`, but it is no longer called. Please retain this validation gate so a keyword-prefill change does not silently remove the existing check for mismatched part assignments.

mainwindow.py
```diff
@@ -909,95 +893,112 @@ def select_part(self, *_):
...
+
+    
+
+    def check_order_number(self):
+        """Verify that the JLC order number placeholder is present."""
+        try:
+            # We open with errors="replace because we're only interested in the occurance of the JLC magic string"
+            with open(self.pcbnew.GetBoard().GetFileName(), errors="replace") as f:
+                data = f.read()
+                return "JLCJLCJLCJLC" in data
```
chmorgan, line 980: [P2] Preserve validation of placeholders on the live board

Searching the saved file for a substring loses the previous live-board count and layer checks. In reproductions, a board with two order-number placeholders no longer receives the multiple-placeholder warning, and deleting the only placeholder without saving also suppresses the missing-placeholder warning because the saved file still contains the text. Fabrication uses the in-memory board, so this checks different data from what is exported. Please retain `count_order_number_placeholders()` and its zero/multiple checks.

mainwindow.py
```diff
@@ -909,95 +893,112 @@ def select_part(self, *_):
...
+                    if name in keys_lower:
+                        txt = (v or "").strip()
+                        if txt:
+                            return txt
+            return ""
+
+        # Listes de noms de champs possibles (ajoutez vos variantes si besoin)
+        keys_partnumber = {"part number", "part_number", "pn"}
+        keys_mpn        = {"mpn", "mfr_part_number", "manufacturer_part_number"}
+        keys_lcsc       = {"lcsc", "jlc", "jlcpcb", "jlc_part", "lcsc_part", "lcsc#"}
```
chmorgan, line 943: [P2] Match the LCSC field names already supported by the plugin

The exact alias set ignores populated fields such as `LCSC Part #`, `LCSC Part`, and `JLCPCB Part #`, even though `helpers.get_lcsc_value()` already recognizes those names and `Store.update_from_board()` uses it to populate the LCSC column. With `LCSC Part # = C25804` and no PART NUMBER/MPN, the existing reader returns C25804 but the new prefill returns the value fallback instead. This reproduces with both `GetFields()` and legacy `GetProperties()` providers. Please reuse or align with the existing reader for the third priority.

chmorgan (changes requested):
Reviewed commit `14980d23c0bac488664d20021205c4bd9585d122`.

Requesting changes for the six findings below. The diff includes unrelated reversions to correction management and fabrication validation: a clean installation cannot load the plugin, and fixing the import alone still leaves correction-list population broken. Please preserve the existing correction and export behavior when adding keyword prefill.

Validation: compared the exact PR head with its parent; used two parallel review workstreams for keyword prefill and correction/logging behavior; executed focused reproductions of the actual Python methods with test doubles, including a temporary SQLite correction database, cancellation, conflicting BOM values, and placeholder checks. Normal keyword priority and fallback cases passed, and the changed file passes a Python syntax check. No repository files were changed. No build or KiCad GUI integration test was run.

Informational: model = GPT-6 (Codex); thinking level = ultra

PR #674, opened 2026-01-12
Refactor DB generation to reduce duplication

z2amiller:
I was working with this file to add Extended Preferred parts and do some speedups, and noticed
that there was quite a bit of duplication between the 'old' and 'new'(?) style databases.

It looks like a lot of changes, but I guess the github PR diff doesn't do move detection so it's
harder to tell that most of this is just shuffling some things around. The substantial changes are:

Changing the sqlite3 rowfactory to sqlite3.row - this makes it return (and accept) a dict with the column names as keys, rather than just having to remember what order you did your 'select' request. It's more verbose but hopefully less mistake prone and more self-documenting.
Moving the iteration of the source database into the base class, and having each subclass implement a row translation method that maps the incoming row from source into the destination row. I don't know enough about the history of this tool to know why there's a parts.db and parts-fts5 database. They seem to have the same columns, so it would simplify this further if it was possible to do the same row translation for each one (e.g. the FTS5 version does some more massaging of prices and descriptions, but maybe that's desirable for the older parts.db as well? LMK and I'm happy to make that change)
Moving the build steps into the base class with an optional per-class override to do post-build steps (e.g. the FTS5 version does the component table and optimization steps)
Using the click progress bar to track the import process rather than printing a bunch of messages every 100k lines.
As far as I can tell this creates more-or-less identical output on my machine. I did some diffs and noticed that the original had some extra spaces in descriptions but that was the only change I could see.

Output spam before: https://pastebin.com/ZZA3eJ9D
Output spam after: https://pastebin.com/qw9ux3nw

(Sorry for the extra work - I was getting annoyed at the merge/rebase and just decided to squash the merge on a new branch)

### DIFF
--- a/db_build/jlcparts_db_convert.py
+++ b/db_build/jlcparts_db_convert.py
@@ -241,9 +241,9 @@ class Generate:
 
                 # The column names have spaces, so map them to placeholders without spaces
                 data = rows[0]
-                columns = ", ".join([f'"{k}"' for k in data.keys()])
+                columns = ", ".join([f'"{k}"' for k in data])
                 placeholders = ", ".join(
-                    [f":{k.replace(' ', '_').replace('.', '_')}" for k in data.keys()]
+                    [f":{k.replace(' ', '_').replace('.', '_')}" for k in data]
                 )
                 newrows = [
                     {k.replace(" ", "_").replace(".", "_"): v for k, v in row.items()}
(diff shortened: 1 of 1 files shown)


z2amiller:
Great, thanks for the confirmation!   Should I keep this PR as-is and then have a second one to clean up the old parts.db file and merge the classes, or would you like me to clean that up as part of this PR?   Maybe a better way to stack it will be to abandon this PR and remove the support for the parts.db first as a new PR, and then stack the refactoring changes on top of that since a lot of the refactoring becomes moot if there's no need to support two mechanisms.  I'll play around with it this week and see what the changes to remove the support for the old DB look like in isolation.

PR #677, opened 2026-01-13
Use the sqlite3 rowfactory in load_tables

z2amiller:
This makes the code much more self-documenting, instead of having to remember the order you selected the columns, they're just in a dict with the column names as keys.


Bouni:
Awesome! Thank you so much for your work!

PR #712, opened 2026-03-28
Make the 'select alike' feature a toggle.

z2amiller:
This makes the  'select alike' feature a toggle,  rather than making it a button-press-per-selection.
When this is selected, any time you select an item it should select all matching items.


Bouni:
Thank you so much for all the work you put in this project!

PR #726, opened 2026-04-10
Project/ipc support

z2amiller:


chmorgan:
Is this for the IPC that is coming or was added to kicad 10? I thought that uses protocolbufs?

z2amiller:
> Is this for the IPC that is coming or was added to kicad 10? I thought that uses protocolbufs?

Yeah this was a first pass at the IPC - with so many recent changes it'd probably need to be done over, which is fine.

I got this working but TBH it's really annoying.  The IPC installations are a different process and managing the plugins is different.   It can run pip which is awesome but that can also mean that the first launch is super slow and looks like it's broken.  You have to explicitly enable IPC support in some random deep settings menu and if you don't, it silently doesn't work. 

To support this, we'd probably just have to draw a line in the sand and park the SWIG version of the plugin on a branch and only port critical bugfixes, and move all development to the IPC version because it will effectively be a completely new package - I think it'll be hard to support both long term.  I'll do another round of research on this.

PR #733, opened 2026-04-13
Add new directories to PCM creation.

z2amiller:
I think this is the right way to fix the archive, but I haven't tested it.


ppi-repo:
i just reinstalled the plugin. the icon is still not showing

z2amiller:
I don't think the plugin archive has been re-created with this fix yet.
(Hopefully tonight!)  This evening I will have some time to test locally
to make sure it'll succeed once BounI@ hits the release button again.

PR #737, opened 2026-04-14
Clean up the PCM archival command.

z2amiller:
- Add a few safety checks
 - Do some cleanup before archive (remove tests and caches and such)
 - Remove some duplicate code and replace with loops
 - Print to stdout if no github_env

I tested this (and PR #733) by creating an archive locally and installing it via the
'Install from File' workflow in the Kicad PCM manager.

### DIFF
--- a/PCM/create_pcm_archive.sh
+++ b/PCM/create_pcm_archive.sh
@@ -1,52 +1,81 @@
 #!/bin/sh
 
 # heavily inspired by https://github.com/4ms/4ms-kicad-lib/blob/master/PCM/make_archive.sh
+set -eu
+
+if [ $# -lt 1 ]; then
+	echo "Usage: $0 <version>"
+	exit 1
+fi
 
 VERSION=$1
+ARCHIVE_DIR="PCM/archive"
+PLUGINS_DIR="$ARCHIVE_DIR/plugins"
+RESOURCES_DIR="$ARCHIVE_DIR/resources"
+ZIP_FILE="PCM/KiCAD-PCM-$VERSION.zip"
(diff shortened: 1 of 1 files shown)


PCM/create_pcm_archive.sh
```diff
@@ -1,52 +1,81 @@
...
+fi
 
 VERSION=$1
+ARCHIVE_DIR="PCM/archive"
+PLUGINS_DIR="$ARCHIVE_DIR/plugins"
+RESOURCES_DIR="$ARCHIVE_DIR/resources"
+ZIP_FILE="PCM/KiCAD-PCM-$VERSION.zip"
+METADATA_FILE="$ARCHIVE_DIR/metadata.json"
+
+sed_inplace() {
```
andy9a9, line 18: I know, that it's too late, but what's the benefit of wrapping `sed -i` into function?

z2amiller: While I was debugging it was handy to have the .bak file - I guess I could've reverted that and just done a straight sed -i.

andy9a9: would be nice :)

z2amiller: aha, I remember now, it's the recommended cross-platform way of doing it.
`sed -i` (no argument) fails on bsd/mac but `sed -i ''` works to create no backup.
but on linux, `sed -i`  works with no backup, but `sed -i ''` fails.  So the cross-platform
thing to do is to `sed -i.bak` which works cross-platform, and then remove the .bak file.

PR #738, opened 2026-04-14
Add step-by-step feedback during fabrication generation

z2amiller:
I ran the new DRC check on a fairly busy / complex board and it took about 8 seconds.  Unfortunately
that hangs the UI with no feedback. :-(

Since the generation is a bunch of steps now, I added some more detailed feedback about the generation
steps.  Getting the UI to update in time to see the message about starting the DRC required some force-flushing
hacks, else the UI doesn't show the 'Starting DRC .... ' message and it just pinwheels.  Unfortunately there's no
callback/feedback from the DRC process itself so there's no way to show progress to the user, but this at least
users know what's hogging the UI thread.

This looks like a lot of diffs but it's a lot of whitespace and/or adding wrappers around the existing calls
in generate_fabrication_data().  The real diffs are in the helpers above.

The number of actual changes should be pretty self-contained.  Tests OK on my large-for-me board
(~13cm x 10cm, ~250 parts)

PR #740, opened 2026-04-14
Switch forced DRC to SWIG API 

z2amiller:
I found a SWIG implementation of the DRC check:

https://gitlab.com/kicad/code/kicad/-/commit/8f01771138fd2517b1cb0fce9b13c20e32de5cd8

I switched to that instead of the CLI to avoid some potential path resolution pitfalls.  I think the CLI based version
was failing for some folks when there were multiple versions of kicad, it could resolve the incorrect version
of the CLI relative to the version of the board.  Unfortunately I don't see a good way to resolve the kicad-cli
associated with the currently running version of the tool, so there'll always be the risk of launching the wrong
CLI.  (The IPC plugin API has a call to resolve the correct binary but that doesn't appear to exist for the SWIG
API)

The SWIG version of the DRC s much faster anyway, since it doesn't need to spin up the whole CLI and load
all of the metadata since that's already in memory.  Unfortunately the SWIG version doesn't output JSON
so the parsing is more complicated but this seems to work ok.


Bouni:
Does this PR require #738 to be merged first?

z2amiller:
This one first - I set the other PR to draft.  I'll rebase these changes onto the SWIG implementation.
(the SWIG is also much faster, and I think will be more reliable for folks that might use multiple
kicad installations)

PR #775, opened 2026-07-08
Add KiCad 10 design variant support

guitar24t:
The plugin previously read only base footprint state (GetValue, IsDNP, GetAttributes, field text), which in KiCad 10 always reflects the default variant, so the part list, project database, and generated BOM/CPL files ignored the active design variant.

- footprint_helpers: resolve LCSC/Value fields, DNP, and the exclude-from-BOM/POS flags through the KiCad 10 *ForVariant APIs for the board's current variant, with getattr guards so KiCad 8/9 and the standalone stubs keep the old behavior.
- Writes are variant-aware too: BOM/POS toggles flip the footprint's variant override (created on demand, deleted again when it ends up mirroring the base so inheritance is restored); LCSC assignment updates an existing lcsc-named override - including one explicitly cleared or holding a placeholder - and otherwise writes the base field, which variants without an override inherit.
- store: sync variant-resolved values into the project database and clear the stored LCSC number when the active variant explicitly blanks the part (lcsc_override_cleared).
- mainwindow: re-sync the store and part list when the active variant changes (on window activation and before Generate), show the active variant in the window title, write the board LCSC field from the paste and mapping-search flows like the part selector does, and warn before exporting LCSC numbers to the variant-independent schematic fields while a variant is active.
- fabrication: match rotation/position corrections against the variant-resolved value and log the same.
- tests: variant helper test suite with fakes mirroring the probed KiCad 10.0.3 semantics (base fallback, flag snapshot on override creation, sparse field overrides, explicit empty override shadowing).

### DIFF
--- a/common/test_footprint_helpers_variants.py
+++ b/common/test_footprint_helpers_variants.py
@@ -1,0 +1,637 @@
+"""Tests for design-variant-aware footprint helpers.
+
+The fakes below mirror the KiCad 10 variant API semantics (verified against
+KiCad 10.0.3): ``*ForVariant`` getters fall back to the base state when the
+footprint has no override object for the variant (or when the variant name
+is empty), field overrides are sparse with per-field fallback, and creating
+a footprint variant override snapshots the base flags at creation time.
+"""
+
+import importlib.util
+from pathlib import Path
+import sys
+import types
+
+ROOT = Path(__file__).parent.parent
+PACKAGE = "kicad_jlcpcb_tools"
+
+if PACKAGE not in sys.modules:
+    pkg = types.ModuleType(PACKAGE)
+    pkg.__path__ = [str(ROOT)]
+    sys.modules[PACKAGE] = pkg
+
+
+def _load_root_module(name):
+    """Load a root module as part of a synthetic package for relative imports."""
+    module_name = f"{PACKAGE}.{name}"
+    if module_name in sys.modules:
+        return sys.modules[module_name]
+
+    spec = importlib.util.spec_from_file_location(module_name, str(ROOT / f"{name}.py"))
+    if spec is None or spec.loader is None:
+        raise RuntimeError(f"Unable to load module spec for {module_name}")
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[module_name] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+helpers = _load_root_module("footprint_helpers")
+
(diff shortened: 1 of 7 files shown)


chmorgan (changes requested):
Review prepared by **OpenAI Codex**, using model **`gpt-6-astra`** with thinking effort **`ultra`**.

Reviewed at `b9086064e026b9e8bdb29d91222a9a2f00fdebff`. Requesting changes for these three findings:

1. **[P1] LCSC assignments leak between variants** — [store.py:422](https://github.com/guitar24t/kicad-jlcpcb-tools/blob/b9086064e026b9e8bdb29d91222a9a2f00fdebff/store.py#L422)

   With an empty default LCSC field and variant A overriding it to `C999`, switching default → A → default leaves `C999` in the database when value, footprint, and exclusions match. The generated default BOM consequently exports A’s part number. The new clearing condition only handles explicit variant overrides. Cached assignments need variant awareness so switching back restores the correct assignment, including an empty one.

2. **[P2] Variant switches retain the previous part’s assembly metadata** — [mainwindow.py:765–766](https://github.com/guitar24t/kicad-jlcpcb-tools/blob/b9086064e026b9e8bdb29d91222a9a2f00fdebff/mainwindow.py#L765-L766)

   Updating the LCSC number through `update_from_board()` preserves cached `assembly_process` and `component_product_type`. Switching from an enriched Standard-only part to an Economic-compatible part therefore incorrectly keeps Standard mode and its fees. Clear and reload metadata when the resolved LCSC changes, and discard pending enrichment results for the previous assignment.

3. **[P2] Exclusion toggles delete preexisting variant overrides** — [footprint_helpers.py:225](https://github.com/guitar24t/kicad-jlcpcb-tools/blob/b9086064e026b9e8bdb29d91222a9a2f00fdebff/footprint_helpers.py#L225)

   Cleanup deletes an existing override whenever its flags match the default. For example, with default flags all false, variant A explicitly has `DNP=False, BOM=True, POS=False`; toggling BOM off deletes that override. A later default DNP change then also changes A’s effective DNP. Preserve preexisting overrides and limit round-trip cleanup to overrides created by the plugin.

**Validation:** All 39 new helper tests passed when executed in memory. Focused in-memory SQLite/export checks and KiCad 10.0.6 API probes reproduced these issues. These scenarios need integration regression coverage. No build or full GUI validation was performed.

PR #781, opened 2026-07-27
Single sided boards get a paste stencil and lose their bottom silkscreen

maximdeclercq:
The 1-layer case slices the bottom list:

```
plot_plan = plot_plan_top + plot_plan_bottom[-2:]
```

`[-2:]` lands on EdgeCuts and PasteBottom, so single-sided boards get a paste stencil for a side with no copper and no bottom soldermask. It also silently depends on the order of a list it doesn't own.

This pull-request should fix that. Also gave the 1-layer plan MaskBottom instead of PasteBottom, a side with no
copper has nothing to paste but is still masked.  Happy to drop it and emit only the top layers plus edges if that's not what JLC expects.

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -312,19 +312,26 @@ class Fabrication:
             ("MaskTop", F_Mask, "Mask top"),
             ("PasteTop", F_Paste, "Paste top"),
         ]
+        # Silkscreen needs no copper under it, so JLCPCB prints a bottom legend
+        # even on a board with a single copper layer.
+        # https://jlcpcb.com/blog/single-sided-pcb-design
+        plot_plan_bottom_silk = ("SilkBottom", B_SilkS, "Silk bottom")
         plot_plan_bottom = [
             ("CuBottom", B_Cu, "Bottom layer"),
-            ("SilkBottom", B_SilkS, "Silk bottom"),
+            plot_plan_bottom_silk,
             ("MaskBottom", B_Mask, "Mask bottom"),
-            ("EdgeCuts", Edge_Cuts, "Edges"),
             ("PasteBottom", B_Paste, "Paste bottom"),
         ]
+        # Edge cuts are not a side, so they are added to every plan.
+        plot_plan_edges = [("EdgeCuts", Edge_Cuts, "Edges")]
 
         plot_plan = []
 
-        # Single sided PCB
+        # Single sided PCB, so bottom copper, mask and paste describe a side that
+        # is not manufactured. JLCPCB publishes no layer set for a 1-layer order,
+        # so this follows from the board rather than from their spec.
         if layer_count == 1:
-            plot_plan = plot_plan_top + plot_plan_bottom[-2:]
+            plot_plan = plot_plan_top + [plot_plan_bottom_silk]
         # Double sided PCB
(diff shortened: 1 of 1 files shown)


fabrication.py
```diff
@@ -316,15 +316,16 @@ def generate_geber(self, layer_count=None):
...
         ]
+        # Edge cuts are not a side, so they are added to every plan.
+        plot_plan_edges = [("EdgeCuts", Edge_Cuts, "Edges")]
 
         plot_plan = []
 
         # Single sided PCB
         if layer_count == 1:
-            plot_plan = plot_plan_top + plot_plan_bottom[-2:]
+            plot_plan = plot_plan_top + [("MaskBottom", B_Mask, "Mask bottom")]
```
chmorgan: AI had some interesting feedback here and I wanted to get your thoughts on:

Finding
- [P2] Keep solder mask on the copper side only — [fabrication.py:328](https://github.com/Bouni/kicad-jlcpcb-tools/blob/547fc637475d2051718f5380a501210983ebbc90/fabrication.py#L328)

  plot_plan_top already contains F_Mask, so adding B_Mask emits solder-mask Gerbers for both sides. KiCad’s through-hole pads create openings on both mask layers by default, meaning B_Mask can contain real apertures even without B_Cu ([KiCad documentation](https://docs.kicad.org/10.0/en/pcbnew/pcbnew.html)). JLCPCB states that its 1-layer service normally has mask on one side and that mask on both sides requires a 2-layer order ([JLCPCB guidance](https://cart.jlcpcb.com/quote?plateType=1)).

  Moving EdgeCuts out of the bottom-layer list and omitting B_Paste are correct, but this branch should be:
  plot_plan = plot_plan_top
  EdgeCuts will still be appended afterward.

maximdeclercq:
Agreed, changed to `plot_plan = plot_plan_top`.

The KiCad half is right, a 1-layer board with one THT pad does plot a
non-empty B_Mask. But JLC describes a single sided board as one side of copper
and one side of solder mask, so that file describes a side that never gets
mask.

Note the EdgeCuts extraction reorders it after PasteBottom for 2-layer and
multilayer boards.

chmorgan:
@maximdeclercq is there a citation on the jlcpcb website we can link to and text we can copy into a code comment to explain? That would provide some traceability into the single sided behavior (I never do single sided boards so its new to me)

maximdeclercq:
@chmorgan I was initially looking at [this blogpost](https://jlcpcb.com/blog/single-sided-pcb-design) and the layers written there, but there is no official Gerber guideline. Either way this should not ruin a board since the realistic downside of getting this wrong is a query and a delay, and to me this is an improvement since there are no excess layers in the zip.

chmorgan:
As long as we have some note in the code as to why we are doing what we are
doing that should provide guidance to the next person editing that code.
You may be the only one expert in that area and even if the git has the
same info in its body, which it should probably have, the code comment is
what will provide that insight to the next engineer.

maximdeclercq:
@chmorgan I have added a comment that describes this. I also found that bottom Silk was skipped in the current `main`, but this should be included according tot the JLCPCB single layer board description I've sent above.

PR #783, opened 2026-07-27
Copper zones that pour no copper are exported silently

maximdeclercq:
`fill_zones` refills and moves on:

```
filler = ZONE_FILLER(self.board)
zones = self.board.Zones()
filler.Fill(zones)      # bool, dropped
Refresh()
```

So a zone that pours to nothing goes to the plotter unfilled, with no logs, and looks the same as a good export.

Fill() cannot report it, since it only returns false on paths this plugin does not reach, and SetIsFilled(true) is set on every non rule area zone whatever the result. GetFilledArea() does not work either, since it sums the layers and hides a zone that fills on F.Cu but comes out empty on B.Cu. So the check is the filled polygon area per layer.

### DIFF
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -1575,10 +1575,33 @@ class JLCPCBTools(wx.Dialog):
                         )
                         return
 
-            self.run_generation_step(
-                "Filling copper zones",
+            refill = self.settings.get("gerber", {}).get("fill_zones", True)
+            empty_pours = self.run_generation_step(
+                "Filling copper zones" if refill else "Checking copper zone fills",
                 self.fabrication.fill_zones,
             )
+            if empty_pours:
+                listed = "\n".join(f"  {pour}" for pour in empty_pours)
+                dialog = wx.MessageDialog(
+                    self,
+                    f"These copper zones contain no filled copper:\n\n{listed}\n\n"
+                    "Plotting now ships the board without that copper.",
+                    "Empty copper zones",
+                    wx.YES_NO | wx.NO_DEFAULT | wx.ICON_WARNING | wx.CENTER,
+                )
+                try:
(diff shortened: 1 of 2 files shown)


mainwindow.py
```diff
@@ -1571,10 +1571,21 @@ def generate_fabrication_data(self, *_):
...
                         return
 
-            self.run_generation_step(
+            zone_problems = self.run_generation_step(
                 "Filling copper zones",
                 self.fabrication.fill_zones,
             )
+            if zone_problems:
+                result = wx.MessageBox(
+                    "Copper zones did not refill cleanly:\n\n"
```
chmorgan: could you explain why the user might want to continue even though the zone filler reported filling issues? wouldn't we instead want to stop and require them to fix the issue before enabling plotting?

chmorgan (changes requested):
Thank you for the contribution, this is a good catch to avoid people a lot of pain when their boards don't come back correctly.

Had a question about the UX that I was hoping you could answer.

maximdeclercq:
My check also had an issue. Fill(aZones, aCheck=false, aParent=nullptr) has no unguarded `return false`, so it always returns true, and zone_filler.cpp:953 calls SetIsFilled(true) on every non rule area zone without looking at the polygon. I changed it such that it now reads GetFilledPolysList(layer).Area() per layer, since GetFilledArea() sums the layer set and hides a zone knocked out on one layer only. This should work normally.

An empty pour is not always a mistake I think, so I would not hard gate this. A zone removed out by a keepout, a leftover 2-corner zone, or a zone on a net with no pads all pour nothing but shoulld be fine. A hard stop would make those boards unexportable.

Also split the fill_zones setting so it gates only the refill, not the check. Someone who manages fills manually still wants to know a pour came out empty.

chmorgan (approved):

PR #784, opened 2026-08-14
Add optional schematic PDF export

RolandWa:
## Summary

This PR is rebased on the current upstream `main` and now contains one focused feature: optional schematic PDF export during fabrication-file generation.

## Changes

- Adds an `Export schematic PDF` setting under Gerber options, disabled by default.
- Runs `kicad-cli sch export pdf` when the setting is enabled.
- Writes the generated PDF to `jlcpcb/production_files/` as:

  `Schematic-<project-name>.pdf`

- Exposes the PDF path to post-generation hooks through:

  `JLCPCB_ARTIFACT_SCHEMATIC_PDF`

- Adds contributor guidance in `.github/copilot-instructions.md`.

## Notes

The earlier broad changes to BOM cost estimation, row highlighting, part assignment, and part-search behavior were intentionally removed from this PR because equivalent or newer implementations now exist in upstream `main`.

## Validation

- `python -m compileall -q fabrication.py mainwindow.py settings.py`
- `git diff --check`

The full test suite and Ruff were not run locally because the development dependencies (`cachetools` and `ruff`) are not installed in this environment.

## Manual Test Plan

1. Open a KiCad project that includes a `.kicad_sch` schematic.
2. Enable **Export schematic PDF** in plugin settings.
3. Run **Generate fabrication data**.
4. Confirm that `jlcpcb/production_files/Schematic-<project-name>.pdf` is created.
5. Confirm that the PDF path is available to a configured post-generation hook through `JLCPCB_ARTIFACT_SCHEMATIC_PDF`.


RolandWa:
## Update: rebased and split into focused commits

I rebased this branch onto the current upstream `main` and rebuilt the changes as small, reviewable commits. The previous broad BOM-estimation implementation was removed because upstream now has a newer dedicated BOM estimator and related UI.

### Included changes

1. **Optional schematic PDF export**
   - Adds a default-off setting to export the project schematic with `kicad-cli`.
   - Writes the PDF to `jlcpcb/production_files/`.
   - Makes the PDF path available to post-generation hooks.

2. **Optional PCB layer PDF export**
   - Adds a default-off setting to export one PDF file for every plotted PCB layer.
   - Files are written to `jlcpcb/production_files/pcb_layers/`.
   - Does not add third-party PDF dependencies.

3. **Part-list status and sorting**
   - In-stock parts are shown before out-of-stock parts, with stock quantity used as a secondary ordering criterion.
   - Rows are color-coded:
     - Grey: excluded from BOM
     - Red: missing LCSC assignment
     - Amber: low stock, below 100
     - Green: high stock, above 10,000
   - Adds a color-coded component summary beside the BOM estimate:
     - Assigned LCSC parts
     - Missing LCSC assignments
     - Parts excluded from BOM

4. **Bulk LCSC assignment**
   - Adds a right-click command at the top of the component context menu:
     `Assign LCSC to matching value and footprint`
   - Requires confirmation and updates the board, project store, UI, enrichment data, and BOM estimate.

5. **Automatic part-search suggestions**
   - Adds an enabled-by-default setting to generate selector search data from the reference designator and component value.
   - Supports R, C, D, Q, and U category suggestions.
   - Preserves slash- and underscore-separated ratings:
     - `100nF/16V` becomes keyword `100nF 16V`
     - `200R/20W/250V` becomes keyword `200Ω 20W 250V`
   - Uses the detected footprint package as the dedicated Package filter, for example `0603`, rather than adding it to free-text keywords.

### Validation

- Focused keyword-normalization tests pass: `3 passed`.
- Changed Python modules compile successfully.
- `git diff --check` passes.
- The broader pytest suite could not run locally because the environment lacks the `cachetools` dependency.

### Notes

All optional or potentially opinionated behavior is behind settings where appropriate:
- schematic PDF export: disabled by default
- PCB layer PDF export: disabled by default
- automatic selector keywords: enabled by default and can be disabled

I also tested the updated plugin copies in KiCad 9.0 and KiCad 10.0 on Windows.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,79 @@
+# Project Guidelines
```
chmorgan, line 1: should this be in AGENTS.md instead? I'm also not convinced we shouldn't instead tell the agent to read the README.md for instructions on building, linting, testing etc rather than duplicating it in here. Any changes to that file will make this file get out of date. I'm not sure anything here isn't overlapping with README.md but maybe I missed something.

mainwindow.py
```diff
@@ -1631,6 +1634,12 @@ def generate_fabrication_data(self, *_):
                 self.fabrication.generate_bom,
             )
 
+            if self.settings.get("gerber", {}).get("export_schematic_pdf", False):
```
chmorgan, line 1637: I think we should default pdf output to true when generating a fabrication package.

settings.py
```diff
@@ -942,6 +972,9 @@ def load_settings(self):
         self.update_lcsc_bom_cpl(
             self.parent.settings.get("gerber", {}).get("lcsc_bom_cpl", True)
         )
+        self.update_export_schematic_pdf(
```
chmorgan, line 975: and schematic to true as well. Not sure I see the cost and it's a huge benefit later. I've often found myself wishing I had it as a part of the production package.

fabrication.py
chmorgan, line 1: I like this change. Possible to spilt it out as its own PR?

datamodel.py
chmorgan, line 1: do you have a screenshot of the UI with this change?  we want to make sure if coloring rows to make said coloring obvious as to what it indicates. I'd also propose splitting this out into its own PR.

mainwindow.py
chmorgan, line 1: I like it. Do you have a screenshot of the popup?

tests/test_search_keywords.py
chmorgan, line 1: this commit looks like a fixup of the previous commit

mainwindow.py
chmorgan, line 1: screenshot of this feature?

chmorgan (changes requested):
thank you for the contribution. I had requests for a couple of screenshots and it would be helpful if these were in separate PRs for separate consideration.

mainwindow.py
RolandWa: image
Component status - easy to find the Assigned, Missing, and Excluded

RolandWa: Speed up the finds of proper components - not one by one, but If w use the one 10k 0402 resistor in hundreds place help a lot.

image

datamodel.py
RolandWa: Easy to see the status of the components :
image

tests/test_search_keywords.py
RolandWa: Test for the Capacitor ratings separated by a slash remain searchable, Resistor value and slash-separated power and voltage remain searchable, Underscore-separated value terms are searchable for ICs

RolandWa:
Thank you for the review feedback. I've split this PR into five focused, independently reviewable PRs as requested:

| # | Feature | PR |
|---|---------|-----|
| A | Optional schematic PDF export | #840 |
| B | Optional per-layer PCB PDF export | #841 |
| C | Color-coded component rows + assignment status summary | #842 |
| D | Bulk LCSC assignment for matching value and footprint | #843 |
| E | Auto-generate part-selector keywords from component values | #844 |

Closing this PR in favour of the above. Screenshots can be added to the individual PRs as needed.

PR #785, opened 2026-08-27
Preserve exclude-from-BOM state through schematic export

### LINKED ISSUE #328: Export to schematics does not preserve BOM/POS exclude flags
I am using this wonderful plugin with KiCad 7.0.6. When I export to schematics, the "Exclude from BOM" and "Exclude from POS" attributes are not exported, so they get overwritten the next time I update the PCB from schematics.

jOaSbA:
Fixes #328

Footprint attributes for "exclude from bill of materials" have a native
counterpart on schematic symbols: the `(in_bom yes|no)` field. KiCad's own
"Update PCB from Schematic" forward-annotates that field onto the
footprint, so leaving it untouched (always the "yes" default) silently
resets the footprint's exclude-from-BOM flag on every schematic update.

This patches the symbol's existing `in_bom` field to match the part's
stored `exclude_from_bom` state, alongside the existing LCSC property
update, across all three supported KiCad schematic formats (v6, v7, v8+).

"Exclude from position files" has no schematic-side equivalent (it's a
PCB/footprint-only concept - position files are about physical placement,
not something a schematic symbol can represent), so it's out of scope
here. That flag being unconditionally wiped on update was in fact a
KiCad-level bug, fixed upstream in KiCad 7.0
(https://gitlab.com/kicad/code/kicad/-/issues/13616), not something this
plugin needed to work around.

Verified against a real fragment of KiCad's own bundled demo project
(RoyalBlue54L-NFC-Antenna), which happens to exhibit the exact mismatch
this fixes - see the two new tests in tests/test_schematicexport.py.

### DIFF
--- a/schematicexport.py
+++ b/schematicexport.py
@@ -1,5 +1,6 @@
 """Module for exporting LCSC data to schematic."""
 
+import glob
 import logging
 import os
 import os.path
@@ -15,16 +16,214 @@ class SchematicExport:
 
     # This only works with KiCad v6/v7/v8 files, if the format changes, this will probably break
 
+    # A symbol's own Reference property, and any per-instance reference from
+    # a reused hierarchical sheet.
+    _TOP_REF_RX = re.compile(r'\(property\s"Reference"\s"([^"]*)"')
+    _INSTANCE_REF_RX = re.compile(r'\(reference\s"([^"]*)"\)')
+    _PROJECT_RX = re.compile(r'\(project\s"([^"]*)"')
+    _UUID_RX = re.compile(r'\(uuid\s"?([^"\s)]*)"?\)')
+    _PATH_RX = re.compile(r'\(path\s"([^"]*)"')
+    _IN_BOM_RX = re.compile(r"\(in_bom\s(yes|no)\)")
+
     def __init__(self, parent):
         self.logger = logging.getLogger(__name__)
         self.parent = parent
+
+    @staticmethod
+    def _is_symbol_start(line):
+        """Return True for the opening line of a placed symbol, in v6, v7 and v8+ files.
+
+        A locally modified symbol carries (lib_name ...) ahead of its
+        (lib_id ...), so keying on lib_id alone silently misses it and
+        shifts every following symbol's data by one. Definitions inside
+        (lib_symbols ...) name their symbol on the opening line itself,
+        so they never match.
+        """
+        stripped = line.strip()
+        return stripped == "(symbol" or stripped.startswith(
+            ("(symbol (lib_id", "(symbol (lib_name")
+        )
+
+    def _project_name(self):
+        """Return the project name KiCad writes into (project "...") instances.
+
+        The board filename is only a guess at it: a board saved under a
+        different name still belongs to the project named by the
+        .kicad_pro file sitting next to it.
+        """
+        found = glob.glob(os.path.join(self.parent.project_path, "*.kicad_pro"))
+        if len(found) == 1:
+            return os.path.splitext(os.path.basename(found[0]))[0]
+        return os.path.splitext(self.parent.board_name)[0]
+
+    def _resolved_bom(self, refs, store_parts):
+        """Return the exclude-from-BOM state shared by every ref, or None when there isn't one."""
+        matched = {p["reference"]: p for p in store_parts if p["reference"] in refs}
+        if not matched:
+            return None
+        missing = refs - matched.keys()
+        if missing:
+            self.logger.warning(
+                "Reused symbol %s is missing PCB data for %s, skipping",
+                sorted(refs),
+                sorted(missing),
+            )
+            return None
+        states = {p["exclude_from_bom"] for p in matched.values()}
+        if len(states) != 1:
+            self.logger.warning(
+                "Reused symbol %s disagrees on its BOM state, skipping", sorted(refs)
+            )
+            return None
+        return states.pop()
+
+    def _placed_symbols(self, lines):
+        """Map each placed symbol to its uuid, its in_bom line and every reference it resolves to."""
+        project_name = self._project_name()
+        symbols = []
+        uuid = None
+        top_ref = None
+        bom_line = None
+        per_project = {}
+        has_instances = False
+        current_project = None
+        started = False
+        symbol_end = None
+
+        def finish():
+            if project_name in per_project:
+                instance_refs = per_project[project_name]
+            elif list(per_project) == [""]:
+                # KiCad writes (project "") for a sheet with no owning
+                # project, so a sole empty name is this project's own.
+                instance_refs = per_project[""]
+            elif per_project:
+                self.logger.warning(
+                    "Reused symbol's instances %s don't include project %s, skipping",
+                    sorted(per_project),
+                    project_name,
+                )
+                instance_refs = set()
+            elif has_instances:
+                self.logger.warning(
+                    "Symbol %s has an empty instances block, skipping", uuid
+                )
+                instance_refs = set()
+            else:
+                instance_refs = None
+            if instance_refs is None:
+                refs = {top_ref} if top_ref else set()
+            else:
+                refs = set(instance_refs)
+            symbols.append({"uuid": uuid, "refs": refs, "bom_line": bom_line})
+
+        for index, line in enumerate(lines):
+            if self._is_symbol_start(line):
+                if started:
+                    finish()
+                started = True
+                uuid = None
+                top_ref = None
+                bom_line = None
+                per_project = {}
+                has_instances = False
+                current_project = None
+                # The symbol's fields end at the closing paren indented
+                # like its opening one. Anything after that belongs to a
+                # sheet or an instance table, not to this symbol.
+                symbol_end = line[: line.index("(symbol")] + ")"
+            if not started:
+                continue
+            if line.rstrip() == symbol_end:
+                finish()
+                started = False
+                continue
+            if uuid is None and (m := self._UUID_RX.search(line)):
+                uuid = m.group(1)
+            if bom_line is None and self._IN_BOM_RX.search(line):
+                bom_line = index
+            if m := self._TOP_REF_RX.search(line):
+                top_ref = m.group(1)
+            if "(instances" in line:
+                has_instances = True
+            if m := self._PROJECT_RX.search(line):
+                current_project = m.group(1)
+                per_project.setdefault(current_project, set())
+            elif current_project is not None and (
+                m := self._INSTANCE_REF_RX.search(line)
+            ):
+                per_project[current_project].add(m.group(1))
+        if started:
+            finish()
+        return symbols
+
+    def _v6_root_instance_refs(self):
+        """Map each symbol uuid to its reused-sheet references, read from KiCad 6's project-wide symbol_instances table.
+
+        Only the project's own root sheet is read. A reused child sheet can
+        carry another project's table, so picking whichever selected file
+        happens to hold one would resolve references against the wrong
+        project.
+        """
+        root = os.path.join(
+            self.parent.project_path, self._project_name() + ".kicad_sch"
+        )
+        try:
+            with open(root, encoding="utf-8") as f:
+                lines = f.readlines()
+        except OSError:
+            lines = []
+
+        refs = {}
+        in_instances = False
+        symbol_uuid = ""
+        for line in lines:
+            if "(symbol_instances" in line:
+                in_instances = True
+                continue
+            if not in_instances:
+                continue
+            if m := self._PATH_RX.search(line):
+                symbol_uuid = m.group(1).rsplit("/", 1)[-1]
+            if symbol_uuid and (m := self._INSTANCE_REF_RX.search(line)):
+                refs.setdefault(symbol_uuid, set()).add(m.group(1))
+                symbol_uuid = ""
+        if not in_instances:
+            self.logger.warning(
+                "No KiCad 6 symbol_instances table in %s; BOM/LCSC sync will be skipped",
+                root,
+            )
+        return refs
+
+    def _sync_bom(self, lines, store_parts, v6_refs=None):
+        """Rewrite the in_bom line of every symbol whose instances agree, in place.
+
+        Runs before the file is read for the LCSC pass, so the two stay
+        independent: this only ever rewrites an existing in_bom value.
+        """
+        for symbol in self._placed_symbols(lines):
+            if symbol["bom_line"] is None:
+                continue
+            refs = (
+                symbol["refs"]
+                if v6_refs is None
+                else v6_refs.get(symbol["uuid"], set())
+            )
+            bom = self._resolved_bom(refs, store_parts)
+            if bom is None:
+                continue
+            desired = "no" if bom else "yes"
+            lines[symbol["bom_line"]] = self._IN_BOM_RX.sub(
+                f"(in_bom {desired})", lines[symbol["bom_line"]]
+            )
 
     def load_schematic(self, paths):
         """Load schematic file."""
         if is_version6(GetBuildVersion()):
             self.logger.info("Kicad 6...")
+            v6_instance_refs = self._v6_root_instance_refs()
             for path in paths:
-                self._update_schematic6(path)
+                self._update_schematic6(path, v6_instance_refs)
         elif is_version7(GetBuildVersion()):
             self.logger.info("Kicad 7...")
             for path in paths:
@@ -34,7 +233,7 @@ class SchematicExport:
             for path in paths:
                 self._update_schematic(path)
 
-    def _update_schematic6(self, path):
+    def _update_schematic6(self, path, v6_instance_refs):
         """Only works with KiCad V6 files."""
         self.logger.info("Reading %s...", path)
         # Regex to look through schematic property, if we hit the pin section without finding a LCSC property, add it
@@ -56,6 +255,8 @@ class SchematicExport:
         newlines = []
         with open(path, encoding="utf-8") as f:
             lines = f.readlines()
+
+        self._sync_bom(lines, store_parts, v6_instance_refs)
 
         if os.path.exists(path + "_old"):
             os.remove(path + "_old")
@@ -133,6 +334,8 @@ class SchematicExport:
         with open(path, encoding="utf-8") as f:
             lines = f.readlines()
 
+        self._sync_bom(lines, store_parts)
+
         if os.path.exists(path + "_old"):
             os.remove(path + "_old")
         os.rename(path, path + "_old")
@@ -205,6 +408,8 @@ class SchematicExport:
         newlines = []
         with open(path, encoding="utf-8") as f:
             lines = f.readlines()
+
+        self._sync_bom(lines, store_parts)
 
         partSection = False
         files_seen = set()  # keeps sheet files already processed.
--- a/tests/test_schematicexport.py
+++ b/tests/test_schematicexport.py
@@ -1,0 +1,1058 @@
+"""Tests for SchematicExport's in_bom/exclude_from_bom sync (issue #328).
+
+Keeps the schematic's in_bom flag in sync with the footprint's
+exclude-from-BOM attribute.
+"""
+
+import importlib.util
+from pathlib import Path
+import sys
+import types
+from unittest.mock import MagicMock, patch
+
+_ROOT = Path(__file__).parent.parent
+
+# Mock KiCad modules before importing schematicexport
+for _mod in ["pcbnew", "wx", "wx.dataview"]:
+    sys.modules[_mod] = MagicMock()
+
+_pkg = types.ModuleType("kicadplugin")
+_pkg.__path__ = [str(_ROOT)]
+sys.modules["kicadplugin"] = _pkg
+
+_version_mod = types.ModuleType("kicadplugin.core.version")
+_version_mod.is_version6 = lambda version: False
+_version_mod.is_version7 = lambda version: False
+_core_pkg = types.ModuleType("kicadplugin.core")
+_core_pkg.__path__ = [str(_ROOT / "core")]
+sys.modules["kicadplugin.core"] = _core_pkg
+sys.modules["kicadplugin.core.version"] = _version_mod
+
+_spec = importlib.util.spec_from_file_location(
+    "kicadplugin.schematicexport", _ROOT / "schematicexport.py"
+)
+assert _spec is not None and _spec.loader is not None
+_se_mod = importlib.util.module_from_spec(_spec)
+_se_mod.__package__ = "kicadplugin"
+sys.modules["kicadplugin.schematicexport"] = _se_mod
+_spec.loader.exec_module(_se_mod)  # type: ignore[union-attr]
+
+SchematicExport = _se_mod.SchematicExport  # type: ignore[attr-defined]
+
+# Trimmed real-world fragment (KiCad 10 / v8+ format) of a single placed
+# symbol, taken from KiCad's own bundled RoyalBlue54L-NFC-Antenna demo. The
+# footprint side of this exact demo has "exclude_from_pos_files
+# exclude_from_bom" set, while the schematic symbol still says "(in_bom
+# yes)" - the mismatch reported in issue #328.
+SCHEMATIC_FRAGMENT = """\
+(kicad_sch
+\t(version 20250114)
+\t(lib_symbols
+\t)
+\t(symbol
+\t\t(lib_id "Connector_Generic:Conn_01x02")
+\t\t(at 156.21 111.76 180)
+\t\t(unit 1)
+\t\t(exclude_from_sim no)
+\t\t(in_bom yes)
+\t\t(on_board yes)
+\t\t(dnp no)
+\t\t(uuid "d5224ac6-3b29-4f27-99e0-c4e878a39680")
+\t\t(property "Reference" "J1"
+\t\t\t(at 156.21 102.87 0)
+\t\t)
+\t\t(property "Value" "Conn_01x02"
+\t\t\t(at 156.21 105.41 0)
+\t\t)
+\t\t(property "LCSC" "C123"
+\t\t\t(at 156.21 111.76 0)
+\t\t)
+\t\t(pin "1"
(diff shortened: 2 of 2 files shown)


chmorgan (changes requested):
Had AI review this since I'm unfamiliar with the consequences of the change. Spent a half hour reviewing code and discussing and I think it might have found an issue with the changes here.

AI slop after this line
==============

Thanks for tackling this—the one-symbol/one-footprint case looks correct.

One edge case needs handling before merging: reused hierarchical sheets have one shared `in_bom` field but may have multiple instance references. Although `store_parts` contains all footprints, this code matches only the symbol’s single `property "Reference"` value and immediately uses that part’s state:

```python
if value == part["reference"]:
    desired = "no" if part["exclude_from_bom"] else "yes"
    newlines[inBomLineIndex] = ...
```

For example, KiCad’s multichannel demo has a shared symbol whose property reference is `RV2`, while its instances are `RV2`, `RV6`, `RV10`, and `RV14`. This makes RV2’s state determine the shared `in_bom` value for all four instances.

Mixed per-instance exclusion states cannot be represented by one schematic `in_bom` field. I suggest collecting the references from the symbol’s `instances` block and:

- Updating `in_bom` when every matched instance has the same exclusion state.
- Leaving it unchanged and logging a warning when the states differ.

A regression test covering a reused symbol with mixed states would also be helpful.

Also, the new tests currently aren’t collected by the PR’s Actions workflow because pytest is invoked only from `core/` and `common/`.

jOaSbA:
Thanks for looking into this.

On the reused-sheet case: you're right, and I went ahead and fixed it. Both LCSC and in_bom now resolve against every reference a symbol maps to (its own Reference property plus any reused-sheet instances), and only write when all matched instances agree. If they don't, it logs a warning and leaves that symbol untouched instead of picking one instance's state arbitrarily. Added two tests covering agreeing and disagreeing reused instances, using a fragment shaped like the multichannel demo you mentioned.

On the test collection point, you're right, and it's bigger than my file: tests.yml only runs pytest from core/ and common/, so the whole tests/ directory (three files that predate this PR) has never run in CI. Worth a separate issue because I don't know how you want the repo to be structured.

jOaSbA:
Hey, I fixed my MR based on your AI's code change. Check this and we will just continue here, easiest for me and you.

chmorgan:
I'd ask that you have a commit for tests and a commit for the logic changes, its easier to review commit by commit in that fashion.

Sol still has some concerns:

# Comparison Report: `feature/safe-bom-sync` vs. PR #785

## Summary

The current `feature/safe-bom-sync` implementation (`8bfa213`) is safer, smaller, and more narrowly scoped than [PR #785](https://github.com/Bouni/kicad-jlcpcb-tools/pull/785) (`82f3116`).

The updated PR correctly handles empty project names, but it still contains an unsafe fallback for unresolved reused-sheet instances. It also changes existing LCSC resolution behavior, which is outside the BOM synchronization fix and increases regression risk.

## Findings

### P1 — Unresolved reused-sheet instances fall back to one reference

PR #785 selects the symbol’s top-level `Reference` whenever the resolved instance set is empty:

```python
refs = (
    set(instance_refs)
    if instance_refs
    else ({top_ref} if top_ref else set())
)
```

See [`schematicexport.py` lines 66–77](https://github.com/Bouni/kicad-jlcpcb-tools/blob/82f3116f37cfc597ce24af10f1ae41f5a551ecdf/schematicexport.py#L66-L77).

An empty instance set can mean either:

1. The symbol is genuinely standalone.
2. The symbol is reused, but its instances could not be resolved safely.

Those cases should not receive the same fallback. For a reused symbol, updating from only `top_ref` can apply one footprint’s BOM state to every instance sharing that schematic symbol.

This remains possible when:

- A KiCad 7/8 schematic has multiple project groups and none matches the active board.
- The chosen project group exists but contains no usable references.
- A KiCad 6 root `symbol_instances` table cannot be found.
- A KiCad 6 symbol UUID is absent from the root instance map.

The feature branch distinguishes a symbol with no instance metadata from one whose instance metadata is present but unresolved. Standalone symbols use their top-level reference; unresolved reused symbols are left unchanged.

### P1 — KiCad 6 warning contradicts the resulting behavior

When no KiCad 6 `symbol_instances` table is found, the PR warns that synchronization will be skipped and returns an empty map:

```python
self.logger.warning(
    "Unable to find KiCad 6 symbol_instances table; "
    "reused-sheet BOM/LCSC sync will be skipped"
)
return {}
```

See [`schematicexport.py` lines 147–150](https://github.com/Bouni/kicad-jlcpcb-tools/blob/82f3116f37cfc597ce24af10f1ae41f5a551ecdf/schematicexport.py#L147-L150).

However, every symbol has already been initialized with its top-level reference. The root references are then unioned into that set:

```python
for symbol in symbols:
    symbol["refs"] |= v6_instance_refs.get(symbol["uuid"], set())
```

See [`schematicexport.py` lines 194–200](https://github.com/Bouni/kicad-jlcpcb-tools/blob/82f3116f37cfc597ce24af10f1ae41f5a551ecdf/schematicexport.py#L194-L200).

Unioning an empty map does not remove `top_ref`, so the writer can still update `in_bom`. This contradicts the warning and can update a reused symbol using incomplete information.

The feature branch instead uses the UUID mapping exclusively for KiCad 6 BOM resolution. If the mapping is unavailable or lacks the symbol UUID, no BOM update is made.

### P1 — KiCad 6 mixes authoritative and potentially stale references

Even when the KiCad 6 root map is available, PR #785 unions those references with the symbol’s saved top-level reference.

For a reused hierarchical sheet, the top-level `Reference` may belong to whichever project last saved the shared sheet. It can therefore be stale or belong to another project.

Including it can cause an otherwise valid update to be skipped because:

- The foreign reference is absent from the PCB store.
- Its BOM state disagrees with the active references.
- Its LCSC value disagrees with the active references.

The root `symbol_instances` UUID mapping is the authoritative source for KiCad 6 and should be used without adding `top_ref`.

### P2 — LCSC behavior is changed unnecessarily

PR #785 resolves both LCSC and BOM state across every reused instance:

```python
def _resolved_lcsc_and_bom(self, refs, store_parts):
```

See [`schematicexport.py` lines 30–54](https://github.com/Bouni/kicad-jlcpcb-tools/blob/82f3116f37cfc597ce24af10f1ae41f5a551ecdf/schematicexport.py#L30-L54).

This changes established behavior in all three schematic writers. When reused instances have different LCSC values—or one instance is missing from the store—the PR can leave the schematic’s old LCSC value unchanged. Before this PR, LCSC followed the symbol’s top-level `Reference`.

That may be a reasonable policy to discuss separately, but it is not required to fix the `in_bom` synchronization problem. Including it makes the patch larger and adds unrelated compatibility risk.

The feature branch applies all-instance consensus only to BOM state and leaves existing LCSC behavior untouched.

## Correctly Resolved Concern

Both implementations now correctly handle an empty project name:

```scheme
(project ""
    ...
)
```

The PR uses `current_project is not None`, so the empty string is treated as a valid project name rather than as “no project.”

The feature branch tests both outcomes:

- Instances agree, so `in_bom` is updated.
- Instances disagree, so `in_bom` is left unchanged.

The PR currently tests only the disagreement case.

## Test and Scope Comparison

| Area | `feature/safe-bom-sync` | PR #785 |
|---|---:|---:|
| Production changes | `+158/-2` | `+214/-14` |
| Test additions | 390 lines | 611 lines |
| Collected feature cases | 21 | 13 |
| Public `load_schematic()` coverage | Yes | Primarily private methods |
| KiCad 6/7/8 behavior matrix | Yes | Individual cases |
| Both `yes → no` and `no → yes` | Yes | No explicit `no → yes` case |
| Variant `in_bom` protection | Yes | No |
| Empty-project agreement and disagreement | Yes | Disagreement only |
| Foreign/stale top-reference behavior | Yes | Partial |
| Multiple-symbol state isolation | Yes | No |
| Real KiCad-derived fixture | No | Yes |

The PR’s principal testing advantage is its real KiCad-derived v8 fixture. The feature branch uses generated fixtures, but covers more behavioral combinations with substantially less duplication.

The PR’s differing-LCSC test is also not conclusive: the schematic starts with `C1`, the top reference resolves to `C1`, and the final assertion expects `C1`. Both “preserve legacy resolution” and “skip the update” satisfy that assertion.

As a compatibility check, the feature branch’s 21 cases were also run against the PR implementation. Thirteen failed because the PR retained `OLD` LCSC values where existing behavior would update from the top reference. The BOM assertions in those cases passed.

## Validation

- `feature/safe-bom-sync` focused tests: **21 passed**
- `feature/safe-bom-sync` complete suite: **352 passed**
- PR #785 focused tests: **13 passed**
- Ruff lint: **passed on both**
- Ruff format check: **passed on both**
- `git diff --check`: **passed on both**

## Recommendation

Use the feature-branch approach as the basis for the implementation:

1. Resolve consensus only for `in_bom`.
2. Preserve the existing LCSC writers and behavior.
3. Use only the authoritative UUID mapping for KiCad 6.
4. Skip updates whenever reused-sheet references cannot be resolved unambiguously.
5. Keep the empty-project fix.
6. Retain the public-entry-point, cross-version test matrix.

It would also be useful to add focused regression cases for:

- A missing KiCad 6 `symbol_instances` table.
- Multiple unmatched KiCad 7/8 project groups.
- A stale KiCad 6 top reference alongside a valid UUID mapping.

Those are test-coverage improvements; no additional production changes appear necessary.

jOaSbA:
Fixed all three P1s. Logic in one commit, tests in the other.

- Ambiguous instances (two or more projects, none of them the board's)
  no longer fall back to the top-level Reference. They count as unsafe
  and log a warning.
- KiCad 6 with no symbol_instances table now really skips, so the
  warning matches what happens.
- KiCad 6 no longer mixes the symbol's own Reference into the root
  table lookup. The table is authoritative.

Revert the logic commit and all three new tests fail. The old
differing-LCSC test could not fail either way: it stored "C1" for RV2,
the value the fixture already had. It stores "C99" now.

On the P2 I can go either way, happy to move the LCSC part to a
follow-up PR if you want the minimal BOM-only change. Why it is in
here: #793 resolves LCSC from lastRef alone, so on a reused sheet
in_bom is safe while LCSC on the same symbol takes one instance's part
number.

Your v6 handling reads the same symbol_instances table now, so both
branches cover the same KiCad versions. Scope is all that is left to
decide.

chmorgan:
Hi @jOaSbA. Maybe this didn't repush? I'm seeing 5 commits here, not logic in one commit and tests in the other.

jOaSbA:
Squashed, it is two commits now.

chmorgan:
what ai model you using for this? I agree that repeated reviews do tend to find new things, which is frustrating. trying to figure out how we can converge more quickly here.

I've split out the LCSC changes on my side into a separate PR in order to reduce scope and improve parity and review here. It looks like your PR does touch some of the LCSC behavior?

5.6 sol ultra is reporting this feedback between my PR and yours:

## Scope comparison: `feature/safe-bom-sync` vs. PR #785

**Verdict:** `feature/safe-bom-sync` is the stronger implementation for the BOM-only scope of [issue #328](https://github.com/Bouni/kicad-jlcpcb-tools/issues/328).

PR #785’s latest head is `29b0692`. Its tree is identical to the previously reviewed `03296e7`; the latest force-push reorganized history but did not change the implementation.

### Remaining BOM issues in PR #785

#### P1: A sole foreign project is accepted as active

[`_placed_symbols()`](https://github.com/Bouni/kicad-jlcpcb-tools/blob/29b06924bca6c904af1de2e7999c1e77d7eaf7ee/schematicexport.py#L66-L83) accepts any sole project group, even when its name does not match the active board.

A shared schematic can retain instance data from another project. If those foreign references happen to match references on the current PCB, PR #785 can update the current project using the foreign instance data.

The feature branch only accepts:

- An exact match for the board-derived project name.
- A sole legacy project group whose name is empty.

All other unresolved project groups produce a safe no-op.

#### P1: An empty `instances` container falls back to the top reference

PR #785 does not record that an explicit empty `(instances ...)` container was present.

Consequently, this:

```scheme
(instances
)
```

is treated the same as a symbol with no instance data. The implementation then falls back to the symbol’s potentially stale top-level reference.

The feature branch distinguishes:

- No `instances` container: use the top reference for a standalone symbol.
- A present but empty `instances` container: instance resolution failed, so skip the update.

#### P1: KiCad 6 trusts an arbitrary selected file

[`_v6_root_instance_refs()`](https://github.com/Bouni/kicad-jlcpcb-tools/blob/29b06924bca6c904af1de2e7999c1e77d7eaf7ee/schematicexport.py#L112-L156) checks the conventional root followed by the user-selected schematic files, accepting the first file containing `symbol_instances`.

Selection identifies files to update; it does not establish which file is the active project root. A reused child can retain a foreign project’s instance table, so this fallback can apply foreign references to the current board.

The feature branch trusts only:

```text
<project_path>/<board-stem>.kicad_sch
```

If that root cannot provide an authoritative table, KiCad 6 BOM synchronization is skipped.

#### P2: The PR parser is not bounded by the placed symbol

PR #785 continues collecting data for a placed symbol until the next placed symbol or the end of the file.

Reference-bearing instance data appearing after the last symbol can consequently be attributed to that symbol.

The feature branch stops parsing each symbol at its indentation-matched closing parenthesis.

### Scope and implementation size

| Branch | Production | Tests | Total churn |
|---|---:|---:|---:|
| `feature/safe-bom-sync` | `+149/-2` | `+439/-0` | 590 lines |
| PR #785 | `+223/-14` | `+764/-0` | 1,001 lines |

PR #785’s totals include its LCSC policy change. The feature branch preserves the existing LCSC behavior and contains only the BOM synchronization needed for issue #328.

The BOM-only branch therefore changes 411 fewer lines.

### Test approach

The feature branch tests the public `load_schematic()` path across KiCad 6, 7, and 8, including:

- Standalone symbols.
- Reused instances that agree.
- Reused instances that disagree.
- Missing PCB instance data.
- Both BOM transition directions.
- Stale top-level references.
- Exact, empty, foreign, and ambiguous project names.
- Explicit empty `instances` containers.
- KiCad 6 root-table authority.
- Variant-local `in_bom` fields.
- Multiple consecutive symbols.

PR #785’s 16 added tests primarily call private writer or helper methods and do not exercise the public dispatch path.

### Intentional KiCad 6 tradeoff

The feature branch does not support an alternate-named KiCad 6 root solely because the user selected it.

This may skip a legitimate update, but it cannot corrupt the schematic. Supporting alternate roots safely would require stronger project identity, such as correlating full schematic instance paths with the paths stored on board footprints.

That broader discovery work should be handled separately.

### Validation

The finalized `feature/safe-bom-sync` branch is at `e60deb9`.

- Focused schematic tests: `29 passed`
- Full test suite: `360 passed`
- Venv Ruff check: passed
- Venv Ruff format check: passed
- `git diff --check`: passed
- Local and remote branch tips match

The branch was force-with-lease pushed.

Only these tracked files changed:

- `schematicexport.py`
- `tests/test_schematicexport.py`

Existing untracked files were left untouched.

jOaSbA:
Right, that's fair. I've taken the LCSC changes out entirely. in_bom is
now written by a pass of its own before the LCSC pass reads the file, so
no LCSC logic changes at all. I checked it by running upstream and this
branch over KiCad's own demo projects with LCSC data in the store and
every differing line is an in_bom line. I am using Claude for decoding/revieuwing.

Two LCSC bugs I ran into while doing that, both already in upstream and
so in both our branches. 

- A symbol without pins inherits the previous symbol's part number. The
  per-symbol state is only cleared in the pin section, which a symbol
  drawn without pins never reaches.
- A locally modified symbol carrying (lib_name ...) before the first
  plain symbol is skipped by the LCSC pass entirely, because partSection
  never turns on for it. 6 of KiCad's 145 demo schematics start with
  such a symbol.

chmorgan:
@jOaSbA I think Claude isn't doing as good of a job here as sol is. It has significantly more lines of code to accomplish the same thing. Perhaps it can do better with a higher thinking level. I don't mean for this to be AI wars though. Sol also found another issue with the project path, tests updated and repushed so you can take a look. Here is the latest report comparing the two approaches.

# Comparison: PR #785 vs. `feature/safe-bom-sync`

## Revisions compared

- PR #785: `da25bf1`
- `feature/safe-bom-sync`: `def16d1`
- Common base: `fa75880`

This comparison is limited to safe BOM synchronization. Reused-sheet LCSC synchronization is intentionally excluded.

## Executive summary

`feature/safe-bom-sync` implements the same core BOM synchronization behavior with substantially fewer lines and broader versioned test coverage.

The primary behavioral difference is project discovery:

- PR #785 trusts a neighboring `.kicad_pro` when it is the only one present.
- `feature/safe-bom-sync` verifies that a candidate `.kicad_pro` represents the project attached to the open board.

The feature branch therefore avoids selecting an unrelated project file and supports renamed boards when multiple project files are present.

No safe-BOM capability implemented by PR #785 was found to be missing from the feature branch.

## Line-count comparison

Counts are limited to `schematicexport.py` and `tests/test_schematicexport.py`.

| Metric | `feature/safe-bom-sync` | PR #785 | Difference |
|---|---:|---:|---:|
| Production diff | +170 / −3 | +207 / −2 | 37 fewer additions |
| Final `schematicexport.py` size | 445 lines | 483 lines | 38 lines smaller |
| Test diff | +570 | +1,058 | 488 lines smaller |
| Nonblank test lines | 512 | 921 | 409 lines smaller |
| Total diff | +740 / −3 | +1,265 / −2 | 525 fewer additions |
| New production helpers | 4 | 6 | 2 fewer helpers |
| Physical test functions | 9 | 23 | Feature uses parametrization |
| Collected test cases | 77 | 23 | 54 more cases |
| Observed line execution | 329/355, 92.7% | 324/360, 90.0% | Feature +2.7 points |

Overall, the feature branch adds approximately:

- 41.5% fewer total lines.
- 17.9% fewer production lines.
- 46.1% fewer test lines.
- 3.35 times as many collected test cases.

The line-execution measurement used Python’s standard tracer against each branch’s own schematic-export tests. It is statement coverage, not branch coverage.

## Production implementation

### Project identity

PR #785 assumes that the only adjacent `.kicad_pro` file owns the open board. That assumption is not guaranteed.

`feature/safe-bom-sync` instead:

1. Enumerates adjacent `.kicad_pro` files as candidates.
2. Gets the project attached to the open board.
3. Asks KiCad for the loaded project associated with each candidate.
4. Accepts a candidate only when its project pointer matches the board’s project pointer.
5. Requires exactly one authenticated match.
6. Otherwise retains the existing board-stem fallback.

This matters when:

- A sole neighboring `.kicad_pro` belongs to another project.
- A renamed board belongs to one of multiple projects in the directory.

PR #785 can select the wrong project in the first case and cannot resolve the renamed board in the second. The feature branch handles both when KiCad exposes project ownership.

### Symbol and instance handling

Both implementations now:

- Recognize normal placed symbols.
- Recognize locally modified symbols where `lib_name` precedes `lib_id`.
- Stop parsing at the symbol’s indentation-matched closing parenthesis.
- Resolve KiCad 7+ project-specific instance groups.
- Resolve KiCad 6 references from the project root’s `symbol_instances`.
- Require every reused reference to exist on the PCB.
- Require every reused instance to agree on `exclude_from_bom`.
- Skip empty, foreign, ambiguous, missing, or disagreeing instance data.
- Rewrite only the existing base `in_bom` value.
- Avoid changing variant-local `in_bom` fields.

### Structure

PR #785 uses six new production helpers and separates symbol parsing from mutation. This provides somewhat more granular helper boundaries and logging.

The feature branch uses four helpers and combines parsing and update selection in `_bom_updates()`. This reduces the production change by 37 added lines.

Both implementations remain targeted line-and-indentation parsers rather than general KiCad S-expression parsers.

## Test-size comparison

| Metric | `feature/safe-bom-sync` | PR #785 |
|---|---:|---:|
| Test-file lines | 570 | 1,058 |
| Nonblank test lines | 512 | 921 |
| Test functions | 9 | 23 |
| Collected cases | 77 | 23 |
| Schematic-export suite result | 77 passed | 23 passed |
| Observed production-line execution | 92.7% | 90.0% |

The feature branch obtains its larger case count through parametrization instead of repeating large schematic fragments and test bodies.

## Version coverage

| Version or path | `feature/safe-bom-sync` | PR #785 |
|---|---:|---:|
| KiCad 6 | 9 explicit cases | 6 cases |
| KiCad 7 | 14 explicit cases | 3 cases |
| KiCad 8 | 18 explicit cases | Generic 8+ cases |
| KiCad 9 | 18 explicit cases | No explicit case |
| KiCad 10 | 18 explicit cases | KiCad-10-format fixtures, but no explicit dispatch case |

Every feature-branch test case reaches the public `load_schematic()` dispatcher.

Most PR #785 tests directly call private version-specific methods. It has only three public-dispatch tests: KiCad 6, KiCad 7, and generic 8+.

## Condition-by-condition coverage

| Condition | `feature/safe-bom-sync` | PR #785 |
|---|---|---|
| Excluded symbol changes `yes → no` | KiCad 6–10 | Generic 8+ |
| Included symbol changes `no → yes` | KiCad 6–10 | Not tested |
| Already-included symbol remains `yes` | Indirectly covered | Direct test |
| Reused instances agree | KiCad 6–10 | KiCad 6, 7, and 8+ |
| Reused instances disagree | KiCad 6–10 | KiCad 6, 7, and 8+ |
| Missing PCB instance | KiCad 6–10 | Generic 8+ |
| Different LCSC values but same BOM state | Tested with LCSC assertion | BOM assertion only |
| Foreign project data | KiCad 6–10 | Generic 8+ |
| Empty project name | Agree and disagree, KiCad 8–10 | Disagree/no-change only |
| Sole foreign project | Two ownership states, KiCad 7–10 | One generic 8+ case |
| Empty `instances` block | Two ownership states, KiCad 7–10 | One generic 8+ case |
| Ambiguous project groups | Two ownership states, KiCad 7–10 | One generic 8+ case |
| Renamed board | Authenticated owner, KiCad 6–10 | Sole-file heuristic, generic 8+ |
| Multiple neighboring projects | Tested | Not tested |
| Null project-pointer equality | Tested | Not applicable |
| Modified symbol with `lib_name` first | Executable, KiCad 6–10 | Fixture exists but is unused |
| Variant-local `in_bom` | KiCad 8–10 | Not tested |
| Consecutive-symbol state isolation | KiCad 8–10 | Not tested |
| Trailing sheet-level `instances` | Not directly tested; implementation handles it | Direct test |
| KiCad 6 foreign selected root | Exercised through public workflow | Foreign file is created but not selected |
| KiCad 6 unquoted UUID | Parser supports it, fixture uses quotes | Directly represented |

## Fixture strategy

PR #785 uses large, manually embedded schematic fragments. Its primary fixture is a trimmed excerpt copied from a bundled KiCad 10 demonstration project.

This provides useful realism, but it is still a text fragment passed to the exporter rather than a complete project opened by KiCad.

The feature branch generates compact fixtures for each serialization format. This makes it easier to:

- Parameterize KiCad 6–10.
- Vary project ownership independently.
- Exercise both synchronization directions.
- Test adversarial instance-resolution conditions.
- Keep the test file significantly smaller.

The tradeoff is that generated fixtures may omit incidental syntax found in real KiCad files.

## PR #785 coverage weaknesses

### Modified-symbol fixture is not executed

PR #785 defines a fixture where `lib_name` precedes `lib_id`, but no test references it. Removing modified-symbol recognition would therefore not fail its suite.

The feature branch executes this condition across KiCad 6–10.

### Empty-project test is not independently discriminating

PR #785 tests only an empty-project disagreement that leaves `in_bom` unchanged. That test could pass if the empty project were ignored completely.

The feature branch pairs the disagreement case with an agreeing case that must update the value.

### KiCad 6 foreign-root test does not reproduce the selected-file path

PR #785 creates a foreign root schematic but does not pass it through the selected-file workflow. It therefore does not fully reproduce the earlier “first selected table wins” behavior.

The feature branch explicitly selects the foreign root alongside the child schematic and verifies that the foreign table is ignored.

## Advantages retained by PR #785

PR #785 has three narrow test advantages:

1. It directly tests a sheet-level `instances` block following the final symbol.
2. Its fixtures include realistic KiCad 6 unquoted UUID syntax.
3. It includes a trimmed KiCad-derived schematic fragment.

The feature implementation already stops at the final symbol’s matching closing parenthesis, so the trailing-sheet test does not expose a production difference.

## Remaining risks

Neither suite runs the complete matrix inside real KiCad 6, 7, 8, 9, and 10 processes.

The feature branch’s project-pointer behavior has been:

- Tested with API fakes across KiCad 6–10.
- Smoke-tested using the installed KiCad 10 runtime.

KiCad 6–9 pointer equality has not been tested in their actual runtimes.

If ownership cannot be authenticated, the feature branch preserves compatibility by using the existing board-stem fallback. That fallback cannot distinguish a same-named foreign project group.

The feature branch also does not catch exceptions raised by `manager.GetProject(path)`, although missing APIs and null project pointers are handled.

## Recommendation

Use `feature/safe-bom-sync` as the implementation baseline.

It provides:

- Equivalent safe-BOM behavior.
- Safer project ownership resolution.
- Explicit KiCad 6–10 coverage.
- More adversarial and discriminating test cases.
- Approximately half the test code.
- A smaller production change.

If additional test hardening is desired, the useful portions to borrow from PR #785 are limited to:

1. A compact trailing sheet-block regression.
2. Unquoted UUID syntax in the generated KiCad 6 fixture.
3. Optionally, one small KiCad-derived fixture as a serialization smoke test.

These additions would not require adopting PR #785’s larger test structure or its sole-`.kicad_pro` project heuristic.

jOaSbA:
@chmorgan. Yeah, yours is the better one, go with #793.

Would be good to get a Co-authored-by for the two things that came out of
this PR. The lib_name symbols, which were getting skipped on about 5% of
the symbols in KiCad's own demos, and reading the project name from the
.kicad_pro instead of the board filename. But the two LCSC bugs I mentioned 
are still in both branches, neither of us touches that code.

chmorgan:
@jOaSbA let me push an update to 793. I'll open a separate PR to address the LCSC bugs as well, we don't want to let those remain

chmorgan:
Thanks @jOaSbA for working through this fix and identifying the locally modified-symbol and project-identity cases. Those contributions are incorporated and credited in #793, which has now merged.

Closing this PR as superseded by #793, as agreed in the discussion.

PR #792, opened 2026-09-01
Feature/run tests from tests workflow

chmorgan:
Clean up tests so we run all of the tests in the tests/ folder, these were previously being excluded from the test runs for no good reason. I did the original implementation and I should have done this more generic and all directories based approach instead. I think maybe it didn't work because we were missing stubs or something, I can't recall.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,9 +15,5 @@ jobs:
         run: pip install pytest
       - name: Install python packages
         run: pip install .
-      - name: Run core tests
-        working-directory: core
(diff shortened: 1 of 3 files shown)


chmorgan:
@Bouni can you give this one a review? I'd like to get it in so we've got proper test activation. It's pretty short.

Bouni:
Sorry for the delay, I'm very busy at the moment. Feel free to merge changes as long as you've tested them yourself.
Anything that brings the project forward is a win in my opinion 👍

PR #793, opened 2026-09-02
Feature/safe bom sync

chmorgan:
Addresses the BOM portion of #328. Position-file exclusion remains a follow-up.

Based on #785 by @jOaSbA.

## Attribution

That work identified two edge cases incorporated here:

- locally modified symbols can place `lib_name` before `lib_id`
- project identity must come from the matching `.kicad_pro`, rather than being inferred from the board filename

Their contribution is recorded with a `Co-authored-by` trailer on the corresponding implementation commit.

## Behavior

This branch synchronizes a schematic symbol's `in_bom` value only when all of its resolved instances are present on the current PCB and agree on the BOM state. It authenticates adjacent `.kicad_pro` candidates against the project attached to the open board.

When authentication is available but the open board has no project identity, or when it finds zero or multiple matches, the exporter warns once per export and leaves project-specific BOM states unchanged. Unscoped symbols continue to synchronize, and the board-stem compatibility fallback remains only for environments without project identity APIs. Existing LCSC behavior is unchanged.

## Verification

- 57 focused cases cover the distinct KiCad 6, KiCad 7, and shared KiCad 8+ serialization paths, including reused symbols, locally modified symbols, ambiguous references, renamed boards, mixed current/stale project data, missing open-board identity, and zero or multiple authenticated project candidates.
- The five missing-identity cases first fail against the prior implementation because it selects a stale board-stem group; they pass after the fail-closed correction. Six multi-file cases also prove that each authentication failure produces one diagnostic per export rather than one per schematic file.
- A compact, hand-authored fixture models the mixed current/stale project condition observed in KiCad's RoyalBlue54L-Feather demo.
- Andy Miller independently verified project-pointer equality with KiCad 10.0.4 and exercised `_bom_updates()` over six real KiCad 9/10 schematics.
- On current `main`: `1579 passed`; Ruff check and format check pass; `git diff --check` passes.

## Remaining gaps and follow-ups

- The full native workflow—rewrite, reopen in Eeschema, and Update PCB from Schematic—has not been exercised end to end.
- The position-file half of #328 is not part of this BOM-focused change. KiCad 10 now serializes schematic-side `in_pos_files`, so that behavior needs separate implementation and native round-trip validation.
- A missing KiCad 6 symbol UUID warning and the pre-existing LCSC `lib_name`-first scanner issue are separate follow-ups.

### DIFF
--- a/schematicexport.py
+++ b/schematicexport.py
@@ -4,6 +4,7 @@ import logging
 import os
 import os.path
 import re
+from typing import Optional
 
 from pcbnew import GetBuildVersion  # pylint: disable=import-error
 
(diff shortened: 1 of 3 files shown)


chmorgan:
@jOaSbA can you take a look at this? codex 5.6 sol didn't think your PR covered all of the cases correctly so I had it generate this both based on your PR and based on kicad operation. I'm not sure how to test it yet but was hoping you could give it a spin and a review.

jOaSbA:
Interesting, I will take a look at it in a couple hours.

chmorgan:
@jOaSbA the intent isn't to overwrite what you proposed, feel free to adjust to correct gaps and repush, I don't have the ability to easily review your changes and test for correctness so next best thing is to lean on AI for a 2nd opinion and 5.6 sol ultra did find some gaps. If you want the report I can generate one for you or your AI to consume.

jOaSbA:
> @jOaSbA the intent isn't to overwrite what you proposed, feel free to adjust to correct gaps and repush, I don't have the ability to easily review your changes and test for correctness so next best thing is to lean on AI for a 2nd opinion and 5.6 sol ultra did find some gaps. If you want the report I can generate one for you or your AI to consume.

I'm fine. I'll just look at the difference in code later. Do you want me to remove the fix in my original mr so that I can push over this mr with the entirely fixed code? Because if we continue on this branch without merging my orinial we will get merge conflicts.

chmorgan:
If this is complete we can always drop the other PR.

chmorgan:
@jOaSbA  updated and ready for re-review

jOaSbA:
@chmorgan Looked at it. Two things.

I ran it against KiCad's own demo, demos/royalblue54L_feather/RoyalBlue54L-Feather.kicad_sch. That file has three different project names in its instance tables: one empty, five saying RoyalBlue54L-Feather and five saying ThingPlus. The .kicad_pro is RoyalBlue54L-Feather, so ThingPlus is just an old name that stuck around in half the file. With the correct project name and a store containing every reference in that file, only 2 of the 7 symbols get synced. With ThingPlus it's 6 of 7.

Skipping those is fine, I don't want it guessing, but the issue is that it says nothing at all. _bom_updates doesn't log anything when refs ends up empty, and _resolved_bom returns None on "not matched" before it gets to either warning. So the user just sees "Added LCSC's to ... (maybe?)" and has no idea 5 of 7 flags were left untouched. My branch had a warning in that exact spot, can you put something like it back?

Second thing is the KiCad 9 and 10 tests. _symbol() only branches on 6 and 7, so 8, 9 and 10 all produce the same text, and is_version6/is_version7 are both stubbed false so they take the same path too. 36 of the 77 tests are the v8 case with a different label on it. They're not wrong, they just don't test what the description says they do. One fixture taken from a real KiCad file would be worth more than the two duplicates, and that's how I ran into the project name thing above.

z2amiller (approved):
Approving.

 Verified the design's key assumption empirically: under KiCad 10.0.4's Python, board.GetProject() and SETTINGS_MANAGER.GetProject(path) return opaque PROJECT * SwigPyObjects that compare equal by pointer across separate proxies (== True, is False), and a non-existent path returns None. So the .kicad_pro authentication in _project_name works as written. I also ran _bom_updates over six real KiCad 9/10 schematics: every placed symbol was found, lib_symbols ignored, instance tables honoured. Ruff clean, 42 tests pass.

One small ask, non-blocking: _project_name falls back to the board stem silently when zero or several .kicad_pro candidates match. When that happens the user only sees the per-symbol "no instances resolve for project X" warnings and has no way to tell that the authenticated lookup was skipped. A single log line at the fallback would make the renamed-board case debuggable.

Two observations for follow-ups, not for this PR:

* The KiCad 6 path skips a symbol silently when its UUID is missing from the root symbol_instances table; the new warning only fires on the 7+ path.  I don't know how much we need to care about KiCad 6 (it's gotta be 5+ years old?) but at least this is a quick fix.
* The three LCSC passes still detect placed symbols the old way ((symbol (lib_id / (lib_id on the next line), so a lib_name-first symbol that is the first placed symbol gets its in_bom synced but its LCSC property skipped. Pre-existing, but the file now carries two definitions of "placed symbol opening".

PR #794, opened 2026-09-03
Feature/one layer gerber tests

chmorgan:


chmorgan:
@maximdeclercq how does the tests commit look here? I had AI whip up some tests against your changes and in general, to confirm the behavior is correct and not broken in the future.

chmorgan:
@maximdeclercq alright, rebased against your latest push.

maximdeclercq:
Pushed another fix. JLCPCB also supports a bottom silk for 1-layer which was not included, even on `main`.

1-layer is now (*_TOP_LAYER_NAMES, "SilkBottom", "EdgeCuts"). (https://jlcpcb.com/blog/single-sided-pcb-design)

I do think the mock layer constants are KiCad 8 numbering, which got reordered in 9.0 (B_Cu is 2 not 31 on pcbnew 10). Maybe pcbnew.IsCopperLayer() could be useful there.

chmorgan:
@maximdeclercq  looks like this is passing now. These tests look good enough? if so I'll merge your branch and then this test branch

maximdeclercq:
I think it's still KiCAD 8 numbering but the tests are honestly fine so you can merge from my part

chmorgan:
@maximdeclercq it did turn up an issue relative to checking for a copper layer. look at the new classify copper layers' commit

PR #796, opened 2026-09-04
Feature/pr 782 stale footprint tests

chmorgan:


chmorgan:
@maximdeclercq had 5.6 sol ultra whip up some tests for your PR. Thoughts?

chmorgan:
@maximdeclercq same offer here, I can have sol fix the implementation so the tests pass

maximdeclercq:
I think sol found a real issue. A graphic zone on F.Mask or F.SilkS is not a rule area, so fill_zones reported it as "poured no copper". Fixed on #783 with `if not IsCopperLayer(layer): continue`, so a rebase should pass the tests.

I'd drop the wording assertions from sol and keep the behaviour ones though.

PR #800, opened 2026-09-05
fix: keep the footprint list still while auto-selecting alike parts

z2amiller:
With "Auto-select alike" enabled, clicking a part selects every row sharing its value and footprint.
The list then scrolled, dragging the row that was clicked out from under the mouse pointer.
On a board with a dozen 10k resistors it jumped by most of a page.

The selection itself is not what scrolls. The list scrolls its *focused* row into view once the click
has been handled, and selecting the alike rows leaves the last of them focused, so the list
chases whichever match sits furthest down the board. Replacing the selection in one call
and then focusing the row that was clicked points that scroll at a row that is already on screen,
so nothing moves. It also leaves keyboard navigation where the user left it.

The actual fix is tiny (just reset the selected item to the item that caused the select-alike
to fire) plus a bit of a performance improvement in using SetSelections rather than the
individual Select calls.

This also adds a bunch of unit tests, using a mocked WX so it can run in the CI tests, it tests
to make sure that the window doesn't jiggle around when selecting.  (To the extent that the mock
can measure it, anyway) The unit tests are the bulk of the changed lines.


z2amiller:
Rebased onto current `main` now that #803 has landed.

These tests no longer stand up their own wx mock, and no longer import one out of `test_mainwindow_stale_footprints` either — that sibling-test import was the shakiest part of this PR, since it made one test module a library for another. They now load `mainwindow.py` through `tests/wx_harness.py`:

```python
mainwindow = load_mainwindow(
    "mainwindow_select_alike_tests",
    wx=wx_stubs(Dialog=type("Dialog", (), {}), NewIdRef=object, PostEvent=lambda *_a: None),
)
```

The file now passes on its own with no other test module collected, which is what the sibling import previously prevented. Production change is unchanged from the original review. All four checks green.

chmorgan (changes requested):
Reviewed head ee373581c52696ff6d72714fd3129c4783c77feb. One behavior regression remains in the focus restoration.

**P2 — Preserve the focused item when a selection event comes from deselecting a row.** At [mainwindow.py:1371](https://github.com/Bouni/kicad-jlcpcb-tools/blob/ee373581c52696ff6d72714fd3129c4783c77feb/mainwindow.py#L1371), `selected_item` comes from `GetSelection()`; it is not necessarily the row the user just clicked.

Reproduction on the GTK behavior path:

1. Enable auto-select alike and select two unrelated footprints A and B, with no other alike matches for A.
2. Scroll until A is offscreen and B is visible.
3. Ctrl-click B to deselect it.

`OnFootprintSelected()` now sees one selection and expands A. Previously A was already selected, so the loop left the cursor and viewport alone. The new `SetCurrentItem(A)` instead moves the keyboard cursor to A and scrolls it into view on GTK. [wxGTK delegates this call to GTK cursor movement](https://github.com/wxWidgets/wxWidgets/blob/v3.2.8/src/gtk/dataview.cpp#L5149-L5171), whose visibility clamping brings an offscreen target into view.

Please preserve the actual focused item across expansion, accounting for deselection events, and add coverage through the selection-event handler. This finding is based on wx/GTK source inspection and modeled execution of the exact production method; native GTK execution is still needed. Native macOS control checks passed and do not establish GTK behavior.

The 11 focused tests pass on this head, but the new `_FakeList` omits `IsSelected`. Running the five new tests with the base revision's `select_alike_parts` substituted gives four `AttributeError` failures and one pass, so those failures do not demonstrate the original scrolling defect. Please add a stateful `IsSelected` and verify that the scrolling regression fails for the intended reason before the fix.

Minor cleanup: Ruff lint passes, but `python -m ruff format --check tests/test_mainwindow_select_alike_scroll.py` reports that the new test file would be reformatted.

z2amiller:
Pushed ae03b4c.

- `select_alike_parts` now reads `GetCurrentItem()` before `SetSelections()` and restores that item afterwards, falling back to `GetSelection()` only when nothing is focused. On wxGTK the cursor follows the ctrl-click, so in your repro the focus goes back to B and A stays where it was; `Select`/`SetSelections` don't move the GTK cursor, so `SetCurrentItem` was the only scroll source on that path. On wxOSX `GetCurrentItem()` derives from `selectedRow`, so with a single selection it resolves to the same row the previous head focused.
- The fake keeps focus and selection apart now (`click(row, ctrl=)`, `GetCurrentItem`, stateful `IsSelected`), and a new test drives the deselect scenario through `OnFootprintSelected`.
- Failure reasons checked: with the base revision's loop substituted, the two viewport tests fail on the `top` assertions (223 != 0, 92 != 150) with no AttributeError; with ee37358's method substituted, only the new deselect test fails.
- Test file is ruff-formatted.

Same caveat as yours: the GTK reasoning is from the 3.2.8 source, not a native GTK run.

chmorgan (changes requested):
Reviewed head `667048b02356f0d1c13e2c112378e2b430ce9000`. The previous focused-versus-selected-row issue is addressed, and the regression tests now fail against the old implementation for the intended reasons. Two GTK regressions remain.

1. **P2 — Bulk reselection changes subsequent Shift-selection** — [mainwindow.py:1669](https://github.com/Bouni/kicad-jlcpcb-tools/blob/667048b02356f0d1c13e2c112378e2b430ce9000/mainwindow.py#L1669).

   On GTK, select unrelated rows 7 and 200, then Ctrl-deselect 200, leaving row 7 without alike matches. `SetSelections` clears and reselects row 7, moving the selection anchor there. Restoring focus to 200 does not restore the anchor while Ctrl is held. Shift-clicking 202 then selects rows **7–202 instead of 200–202**. Previously, the loop skipped the already-selected survivor and preserved anchor 200. Please preserve native selection state when expansion adds nothing, and add coverage for the subsequent Shift-selection interaction.

   Source evidence: [wx bulk selection](https://github.com/wxWidgets/wxWidgets/blob/v3.2.8/src/gtk/dataview.cpp#L5306-L5334), [GTK anchor handling](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeselection.c#L1569-L1602), [Ctrl guard during cursor restoration](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeview.c#L13319-L13331), and [remaining toggle-event processing](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeview.c#L11017-L11041).

2. **P2 — Enabling Auto-select alike can still jump the viewport** — [mainwindow.py:1674](https://github.com/Bouni/kicad-jlcpcb-tools/blob/667048b02356f0d1c13e2c112378e2b430ce9000/mainwindow.py#L1674).

   Select a row with no alike siblings, scroll it offscreen, then enable Auto-select alike. `toggle_select_alike` invokes this method, and the unconditional `SetCurrentItem` brings the saved row back into view on GTK even though focus never changed. The previous singleton path made no selection calls and left the viewport alone. Please restore focus only when necessary and cover enabling the option with an offscreen selection.

   Source evidence: [wx cursor setter](https://github.com/wxWidgets/wxWidgets/blob/v3.2.8/src/gtk/dataview.cpp#L5149-L5171) calls GTK unconditionally; [GTK's setter](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeview.c#L13476-L13482) applies `CLAMP_NODE`, [including when the cursor is unchanged](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeview.c#L13334-L13352), and [scrolls an offscreen target into view](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeview.c#L9655-L9681).

Validation: 13 focused tests passed, Ruff lint and formatting checks passed, and three native macOS list-control probes passed with wxPython 4.3.1 / wxWidgets 3.3.3, including a simulated native click through the production selection handler. Substituting the original selection loop makes the viewport tests fail on their assertions; substituting the earlier selected-row focus restoration makes the Ctrl-deselection test fail.

Both new findings are supported by wxWidgets 3.2.8 / GTK 3.24.51 source inspection and modeled execution through the production handlers. The source-derived GTK model shows the offscreen-toggle viewport changing from row 190 to row 7, and the subsequent Shift-selection growing from 3 rows to 196. **Native GTK execution was not performed; the macOS probes do not establish GTK behavior.**

z2amiller:
`select_alike_parts` now returns before touching the control when every alike row is already selected, which is the case in both of your repros. That restores what the old loop did there: no `SetSelections`, so the GTK anchor stays on the ctrl-clicked row, and no `SetCurrentItem`, so enabling the option can't clamp an offscreen row into view. The fake list now carries the anchor and scrolls on `SetCurrentItem`; the two new tests fail against the previous head with the shift range growing to 7–202 and the viewport jumping to row 7, as your model predicted.

chmorgan (changes requested):
Reviewed head `19cb93448c4689dc092d056f1c4697454ceb7838`. The previous singleton Shift-click-anchor and offscreen-toggle cases are fixed, but the offscreen-toggle regression remains when expansion actually adds alike rows.

**P2 — Avoid scrolling an unchanged offscreen cursor when enabling Auto-select alike** — [mainwindow.py:1682](https://github.com/Bouni/kicad-jlcpcb-tools/blob/19cb93448c4689dc092d056f1c4697454ceb7838/mainwindow.py#L1682).

Reproduction on the GTK behavior path:

1. Disable Auto-select alike and select row 7, with an alike sibling at row 100.
2. Scroll until the viewport starts at row 150, leaving the selection offscreen.
3. Enable Auto-select alike.

The new early return does not apply because row 100 needs to be selected. On GTK, `SetSelections()` leaves the cursor unchanged, but the unconditional `SetCurrentItem(7)` then clamps that offscreen row into view. The original loop expands the selection without moving the viewport. The new offscreen-toggle test uses only `alike_rows=[7]`, so it covers the no-op case but misses this path.

Please avoid invoking the focus setter when the native current item already equals the preserved item, and extend the toggle regression coverage to include unselected alike siblings while retaining the current selection-expansion and focus tests.

Primary-source evidence: [wxGTK selection methods](https://github.com/wxWidgets/wxWidgets/blob/v3.2.8/src/gtk/dataview.cpp#L5306-L5350) do not move the cursor; [the cursor setter](https://github.com/wxWidgets/wxWidgets/blob/v3.2.8/src/gtk/dataview.cpp#L5149-L5171) calls GTK unconditionally. GTK applies [CLAMP_NODE](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeview.c#L13476-L13482), [even when the cursor is unchanged](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeview.c#L13334-L13352), and [scrolls an offscreen target into view](https://github.com/GNOME/gtk/blob/3.24.51/gtk/gtktreeview.c#L9655-L9681).

Validation and limits:

- 32 focused tests pass; Ruff lint and formatting checks pass for both changed files. All four GitHub checks are green.
- Substituting the original selection loop makes the viewport regression tests fail on their intended assertions. Substituting the preceding `667048b` method makes the two newest singleton tests fail, confirming those fixes. Supplemental guard-cleanup, reentrant-event, and persisted-toggle checks pass.
- Three native macOS list-control probes pass with wxPython 4.3.1 / wxWidgets 3.3.3: expansion above and below the viewport, plus a simulated native click through the production selection handler. These use a temporary fixture window and stubbed KiCad dependencies, not a complete KiCad session.
- Two independent reviewers corroborated the remaining case. Exact base/head handlers executed with a source-derived GTK control model preserve viewport row 150 on the base but move it to row 7 on the head for both `[7, 100]` and `[7, 180]`. **Native GTK execution was not performed; this finding is supported by primary-source inspection and modeled execution.**

z2amiller:
Good catch, cross platform stuff is such a pain.

The bug:

The focus is put back only when SetSelections() actually moved it (GetCurrentItem() before vs after). On wxGTK SetSelections never moves the cursor, so this method makes no SetCurrentItem call there at all; wxOSX still gets one. The selected-row fallback is gone. The rows being added go last in the array, so GTK's anchor lands where main's per-row loop leaves it.

Testing:

I set up a docker image and a real GTK install on a raspberry pi to repro the issue.  (Side note - it might be worth it to do a nightly smoke test run on linux/mac/windows runners or something using KiCad docker images to have it drive the UI around, since a bunch of the latest PRs have focused around WXWidget stuff and we've found a lot of bugs)

I also launched this on Windows (Using Crossover, so there might be some fidelity diff here due to the emulation layers).

Native GTK run (wxWidgets 3.2.8, GTK 3.24.49, Xvfb, real X clicks): 8 scenarios × 5 revisions, measuring selection, cursor, top row and shift-click anchor. 3e12239 matches main in all 8. Every earlier finding reproduces on the head it was raised against: ee37358 scrolls to the ctrl-click survivor, 667048b moves the anchor and scrolls on enable, and 19cb934 jumps from row 150 to row 7 with alike rows at 100 or 180. The run also turned up two differences the reviews didn't raise, and both now match main: on 19cb934 a plain click moved the anchor to the clicked row, and a selection without a cursor scrolled on enable.

Native macOS run (KiCad 10.0.4, real plugin window, 238-footprint board, clicks posted to the AppKit event queue): main's loop jumped the list 1279 px on a first-match click. 3e12239 moved 0 px there, and also on a last-match click, which takes the no-focus-call path.

Windows (KiCad 10 under CrossOver, wxWidgets 3.3.2 generic control): no revision scrolls on the click. On main the focus, which is also that control's shift-click anchor, ends on the last alike row, so the next Down arrow jumps the list (row 0 to 69); from 19cb934 on it stays on the clicked row. Wine isn't Windows, but that control's selection and scrolling logic is wx's own.

Tests. The fakes are split into Cocoa and GTK lists. A parametrized GTK test pins selection, cursor, anchor and viewport to main's loop on six paths, with no SetCurrentItem call. 7d7310f stubs wx.Frame, since #829 otherwise breaks collection on current main. Merged with 7299624: 1907 passed, ruff 0.11.4 clean.

chmorgan (approved):
Approved after re-reviewing head `3e1223961f79426d3a5d1d57b018f9baab71b89d`. No new correctness findings.

The previous GTK findings are resolved: no-op expansion preserves the existing Shift-selection anchor, newly added rows retain the original loop's final anchor, and the conditional focus restoration avoids scrolling an unchanged offscreen cursor, including when expansion adds matches. The no-cursor path is also covered.

Validation:

- 22 focused tests and Ruff lint/format checks passed.
- Substituting earlier implementations makes the regression tests fail for the intended reasons.
- Six supplementary probes passed for guard cleanup, reentrant events, PCB selection synchronization, zero-selection callers, and persisted enabling.
- Five native macOS checks passed using the real part-list model and production handlers, including simulated clicks on first and last matches and a singleton. The viewport and selected/current items remained correct on both focus-restoration paths.

Independent reviews covered platform behavior and test coverage. I did not run native GTK or Windows during this review; those conclusions rely on framework source inspection and modeled tests. The native macOS checks used a temporary fixture window, not a complete KiCad session.

PR #804, opened 2026-09-06
Replace numeric assembly classifications with an enum

chmorgan:
## Summary

Follow up on the enum suggestion from Andy during review of #797.

- Add `ComponentProductType(IntEnum)` with named values for Economic and Standard, Economic Only, and Standard Only.
- Return enum members from the existing normalizer and use them in BOM, pricing, and enrichment comparisons.
- Preserve numeric API values, integer storage, and existing handling of unknown classifications. No UI, pricing-policy, or schema changes.

Production changes and tests are in separate commits. The diff is limited to six files: 44 additions and 16 deletions.

## Validation

- All 428 tracked tests passed.
- Ruff lint and formatting checks passed; commit hooks passed.
- Tests cover enum return values, supported inputs, unknown values, and SQLite integer-storage compatibility.
- Independent scope and compatibility review found no issues.

### DIFF
--- a/bom_estimation/assembly_mode.py
+++ b/bom_estimation/assembly_mode.py
@@ -4,7 +4,16 @@
 # pyupgrade: disable
 
 from dataclasses import dataclass, field
+from enum import IntEnum
 from typing import Optional
+
+
+class ComponentProductType(IntEnum):
+    """Observed JLCPCB assembly classifications with their numeric API values."""
+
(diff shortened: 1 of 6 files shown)


z2amiller:
One nitpick  -- in the store.py the magic numbers are still used in one place - probably worth at least a comment that ties it back to the enum (or a one-liner for loop to join the values).  Fine in a followup PR.

Issue #778, opened 2026-07-14
Settings dialog box improvements


kavanoz64:
The options in the settings dialog box change the text based on whether the checkboxes are checked or not. This is a confusing way of presenting the options and in general against the user interface norms. Certainly Kicad does not do this.

When an option is unchecked and I prefer what the setting is saying, I feel the urge to check the box. Because seeing it unchecked tells me that the opposite of the behavior will happen. And when I check the box in that case, it will do the opposite of what I want.

The text should be static and the checkbox should indicate whether the option is enabled or not.

Example:
Tented vias:
Checked: Yes, tented vias are enabled.
Unchecked: No, tented vias are not enabled.

The images can stay and change with the box to show the state of the behavior.
Also it would conform better with Kicad UI if the check boxes are on the right hand side and aligned horizontally as a single column.

Most of the options will be easy to understand like that. One of them can be either a radio button or drop down which is "LCSC number from xxx have priority". It can be changed to say: "LCSC numbers used from " followed by the options.

Example Mockup:

<img width="1072" height="277" alt="Image" src="https://github.com/user-attachments/assets/64a18205-d0b2-40d2-adc2-27048aec8661" />

EvanBarnesCL:
I would love this too! The current behavior is baffling. I came here specifically to figure out how to interpret the settings page, I was really confused.

PR #809, opened 2026-09-06
Automatically remember and apply part preferences

### LINKED ISSUE #802: proposal: either 'Remove parts mapping (preferred parts mapping)' or 'Automatically save parts mapping'
Does parts mapping help? Right now you save all of the mappings when you hit 'Save Mapping' or none of them. You can't pick individual parts to map.

I'd argue that we should either make the save automatic, which lets us remove a button and simplify the workflow, or we should remove the preferred parts mapping entirely.

@Bouni  @z2amiller thoughts?

chmorgan:
Part preferences remember an LCSC part for an exact value and footprint combination across projects. Picker and paste choices are remembered immediately; eligible empty assignments fill once when the plugin window opens. Both options default to enabled, and manual Save/Apply remains available with automation disabled.

- **Remember my part preferences** records successful explicit choices. A later choice replaces the preference; opening or closing a board does not learn its assignments.
- **Parts preferences fill in empty LCSC assignments** runs after board/project reconciliation. Existing assignments and occupied raw fields are preserved; DNP parts and parts excluded from BOM or placement are skipped.
- **Part preferences** names the toolbar, manager, settings section, context actions and CSV dialogs. Tooltips explain timing and exclusions. Settings use `part_preferences.remember_lcsc_assignments` and `part_preferences.fill_empty_lcsc_assignments_on_open`.

Reusable preferences require a complete `C[0-9]+` identifier. Lowercase `c` and surrounding whitespace are accepted and normalized. Invalid new choices are skipped with a warning; invalid installed rows remain visible/exportable but are not applied. Existing board text is preserved by automatic filling regardless of validity. A matching preference blocked by a nonempty LCSC/JLC field logs the reference, field and value.

Clearing an assignment retains its preference, so an eligible blank can fill on the next opening. Existing `mappings.db`, its three-column `mapping` table and the exported CSV format remain compatible. CSV import parses the complete file and checks its structure before saving valid rows in one transaction.

This revision replaces five commits with three and removes **195 net lines: 66 production lines and 129 test lines** compared with `18474c6`. One assignment helper handles reference-to-part choices, with catalog details cached per part. Storage validation lives in one batch API. Tests share real SQLite fixtures while retaining rollback, reopening, native selection lifetime and recovery coverage.

### Commit series

1. `081b90c` — Use Part preferences terminology — existing UI, dialog and controller names and explanations.
2. `dabd16a` — Validate and batch part preference persistence — shared storage, CSV import and persistence regressions.
3. `67d138c` — Remember choices and fill empty assignments safely — settings, unified assignment flow, project transactions, startup recovery and workflow regressions.

#812 has merged. The series is based on upstream `main` at `1110a6b`, including the subsequent correction and settings fixes. **Final review and discussion remain required before merging.**

### Acceptance evidence

| Requirement | Evidence |
| --- | --- |
| Immediate remembering and lifecycle | Real Store/Library controller tests cover picker, stateful clipboard paste, manual Save/Apply, disabled options, later choices, clear/reopen and once-per-window filling. Automatic actions report actual changed-key/assignment counts; unchanged or failed saves emit no success log. |
| Validation and existing data | Storage and controller regressions cover ASCII C-numbers, lowercase/padded input, invalid prefixes, embedded/suffixed IDs, Unicode digits and nonstrings. Installed invalid rows are retained and skipped. Exact quoted/whitespace keys and legacy duplicate rows remain compatible; the first installed duplicate determines lookup until a valid save updates all duplicates. |
| Raw fields, aliases and native state | Opening preserves non-C text, lowercase, trailing/only whitespace and multiple aliases. Explicit C-number assignment and Clear synchronize C-valued aliases while preserving tested rotation/position/layer/URL metadata. Native checks save/reload an actual PCB and verify borrowed wx multi-selection storage stays alive through model updates. |
| Shared transaction and CSV | Real SQLite tests verify one transaction, distinct-key counts, legacy duplicate updates and later-row INSERT/UPDATE rollback through fresh connections. Strict CSV parsing, malformed/header/column/decoding failures and later database failure are covered. A failed shared save retains the committed project choice. |
| Project transaction | Later LCSC or stock failure is injected across picker, paste, manual Apply, Clear and automatic fill. Stored rows, metadata, board fields and displayed rows remain unchanged; no assignment-success event follows the failed transaction. Missing catalog details and stale references are covered. |
| Startup and retry | Real Library construction handles optional preference schema failures and a nonempty database missing its table. Native locked/malformed project databases and failed Library construction leave Settings usable and disable dependent actions. A real directory-picker event repairs Library/Store. Controller tests cover postcommit enrichment read failure, pending work reset, rescheduling and stale-result rejection. |
| Settings and upstream integration | Correction recovery remains covered. Stateful checkbox tests exercise both states and feature-specific keys. Atomic file tests cover enabled upgrade defaults, retained false/custom values, settled loads, failed migration persistence and retry. Native checkbox events persist across reopening. |

Validation: **1,522 tests passed** using the repository's existing `myenv`; repository-wide Ruff, formatting of all 19 changed/new Python files, Python 3.9 syntax checks and `git diff --check` passed. New validation/CSV regressions failed against the previous behavior before their fixes. The first two commits passed their full suites independently: **1,372** and **1,412** tests. Their exact tested trees match the committed patches. Acceptance tracing, an independent requirements-first review and the final patch/commit audit are complete. All 224 final source files match the tested snapshot; no new blocking findings remain.

Native validation passed **135 assertions** using installed KiCad **10.0.6**, wxPython **4.2.2a1** and wxWidgets **3.2.8** on macOS. The harness used real window constructors, controls/events, SQLite files and `pcbnew` board save/reload with an isolated catalog and board provider.

Coverage limits:

- Windows/Linux and the complete plugin-launch lifecycle inside the PCB editor were not exercised. Remote enrichment and initial remote correction seeding were suppressed in the native harness. The generation guard was tested; fabrication generation/builds were not run.
- Transactions are atomic within each SQLite database; shared preferences, project storage and native board/model state do not form one distributed transaction.
- Automatic filling conservatively treats any nonempty LCSC/JLC-prefixed field as occupied. Explicit alias recognition retains the existing C-number heuristic. Clipboard extraction and the legacy board reader/highlighter are unchanged.
- Optional shared preference storage failure is logged while Settings and project assignments remain usable. The Part preferences manager does not recover from a malformed shared database; core project/Library startup failures have the visible recovery state shown below.
- The pre-existing ancient catalog-to-preferences migration defect is outside this change; compatibility here covers installed three-column preference data.

### Screenshots

These earlier native captures show the unchanged interface. Fresh capture for this revision was blocked by the computer-use tool (`cgWindowNotFound`); the 135 native functional assertions above were rerun successfully.

**Settings:** approved labels with details in tooltips.

Part preferences settings

**Main window:** Part preferences toolbar and native assignments. Existing raw non-C fields remain on the board even where the legacy reader leaves their displayed LCSC cell blank.

Part preferences toolbar and assignments

**Manager:** shared preference explanation and stored rows. Legacy non-C rows remain visible; the new validation skips them when applying preferences.

Part preferences manager

**Recovery:** malformed project storage is visible, assignment/generation controls are disabled, and Settings remains accessible.

Recoverable project storage failure

Fixes #802.


z2amiller (commented):
Behavior and defaults look right to me — this is what we landed on in #802, and skipping DNP / exclude-from-BOM / exclude-from-POS gives a workable way to keep a part intentionally blank.

I checked a few of the claims empirically before commenting:

- `python -m pytest -q tests common/test_store_pad_filter.py` → 147 passed; `ruff check .` clean.
- The mapping key written by `_apply_lcsc_assignments` is `str(fp.GetFPID().GetLibItemName())`, which matches what `Store.update_from_board` puts in `part_info.footprint` — so auto-saved keys and auto-loaded / manually-found keys agree.
- `plugin.py:Run()` constructs a fresh `JLCPCBTools` per invocation, so the `_mapping_autoload_done` guard really is per-window-open as the README says.
- `_auto_load_mappings` runs after `Store.__init__` → `update_from_board()`, so "after board reconciliation" is accurate.

**Worth calling out as a real fix:** parameterizing `get_mapping_data` / `delete_mapping_data` / `update_mapping_data`. Since `insert_mapping_data` was already parameterized, a footprint name or value containing an apostrophe could be written to the mappings DB and then never looked up, updated, or deleted again — the f-string lookups would either error or silently miss. That's a live data bug independent of the feature, and it's good to have it closed.

Also good: collapsing the four divergent assignment paths (`assign_parts`, `paste_part_lcsc`, `search_foot_mapping`, auto-load) into `_apply_lcsc_assignments`. Paste previously skipped both `set_lcsc_value` and `store.set_stock`, and used `details["type"]`, which would `KeyError` on a part missing from the catalog. And moving `library.create_mapping_table()` ahead of `init_store()` in `init_data` is the right ordering fix for the new read.

The tests are more than a rubber stamp — `tests/test_mainwindow_mappings.py` loads the real `store.py` and `footprint_helpers.py` against a fake footprint implementing `GetFields` / `SetField` / `GetAttributes` / `IsDNP`, rather than mocking the modules under test.

---

### One ask before merge: log the auto-fill

`_auto_load_mappings` currently logs only on `sqlite3.Error`, so a successful auto-fill is completely silent. A user who opens the plugin and finds LCSC numbers they didn't assign has nothing to search for and no reason to look in Settings. Worth an `info` at the point it happens:

```python
filled = 0
for lcsc, references in references_by_lcsc.items():
    filled += len(self._apply_lcsc_assignments(
        references, lcsc, update_model=False, notify=False
    ))
if filled:
    self.logger.info(
        "Auto-filled %d blank LCSC assignment(s) from saved mappings. "
        "Disable this under Settings > Footprint mappings.",
        filled,
    )
```

`_apply_lcsc_assignments` already returns the list of assigned references, so the count is free.

This matters most for upgrading users: `migrate_settings` turns `auto_load` on for everyone, and existing mappings DBs were populated by the old bulk "Save mappings" button, which was easy to fire accidentally and saved every row at once. A log line gives those users a thread to pull on.

Related: `search_foot_mapping` dropped its `self.logger.info("Found %s", lcsc)` in the refactor — probably worth keeping an equivalent there too.

### Nitpicks (all optional)

- `library.save_mapping_data` returns a bool that `_save_mapping` discards. Either use it or drop the return.
- `images/mdi-content-save-settings.png` is now unreferenced, and `ID_SAVE_MAPPINGS = 15` leaves a gap before `ID_EXPORT_TO_SCHEMATIC = 16`.
- `sanitize_lcsc` now uppercases its result. Harmless — one caller, and `_apply_lcsc_assignments` uppercases anyway — but it's an unflagged behavior change in an otherwise unrelated hunk; worth a line in the PR description.
- The two new checkboxes live in their own `StaticBoxSizer` rather than the two-column `settings_grid` every other setting uses, and use tooltips where the others use the help-button helper. Consistent with `hooks_box_sizer`, so no worse than what's there — just a heads up that #778 is reworking that layout.
- `_auto_load_mappings` wraps `store.read_all()` and the mapping lookups in `except sqlite3.Error`, but the `_apply_lcsc_assignments` loop that follows is outside the `try`. A `store.set_lcsc` failure there would propagate out of `init_store` and the window wouldn't open. Cheap to widen.

chmorgan:
@z2amiller alright, this is a bit more polished. Edited the language, integrated your comments and plan to rebase and merge after 812 lands. Biggest change is switching from the very generic 'Mapping' to 'Part preferences' which I think captures this feature more precisely. But I'd be interested in feedback in any case.

z2amiller:
Big commit here, trying to be careful on review:

1.  Would it make more sense to "commit" the write on gerber generation (or mainwindow close) rather than on every change?  (Just in terms of potentially polluting the global database with potentially transient/inflight changes?).  This isn't a demand but more of an open question.
2. The update to the database should probably be encapsulated a little more - I think right now it's churning the connections / updates to the database a lot (per row).  It'd be more efficient to collect all of the changes and then  pass it to a 'StoreAll' method that can persist the whole thing in one transaction rather than a bunch of roundtrips.
3. Right now, LCSC text that fails to validate gets overwritten with the value from the global database -- I'm not 100% sure if this is the right call or not.  It's probably "principle of least surprise" to have a filled but invalid row not be overwritten . One followup here is that I'd like to highlight things that fail the regexp check (lowercase 'c', trailing space, etc), but that would come later.  (I'm working on some PRs to consolidate how we format and validate LCSC numbers, this could be included in that work)
4. We should have some logged autofill feedback (Autofilled X components from global mappings, toggle this in Settings -> XXX / Saved X components to global mappings / Toggle this in Settings -> XXX) - I know not all users read the log but it at least gives some breadcrumbs about why a bunch of stuff magically appeared.
5. The apply loop at the end of _fill_empty_lcsc_assignments_from_part_preferences is still outside the try, and it runs inside __init__ now, so a locked /invalid project.db (same project open in a second KiCad) would stop the window from opening at all - and the off switch is behind that window. Cheap to widen.
6. Heads up for the rebase: onto current main it merges clean but 62 tests fail, all on the test side. tests/correction_test_support.py and tests/test_library_corrections.py still use mappingsdb_file / create_mapping_table; init_store now reads self.settings and self.logger, which test_mainwindow_correction_recovery._population_window doesn't set; test_settings_persistence seeds a file without part_preferences, which the migration rewrites on load; and your own fixture needs ComponentProductType in the bom_estimation.assembly_mode stub plus correction_status / Layout from #816.

chmorgan:
@z2amiller

1. I guess? It didn't seem expensive to save on each change though, although I get the idea that Gerber generation implies some fitness of the mapping that saving on changes isn't achieving.
2. makes sense
3. note there are a bunch of parts that start with 'z' and other letters even.
4. Makes sense
5. agree
6. let me rebase and repush.

z2amiller:
Thanks for addressing the comments, this looks ready but I want to push back on one thing.
(Or rather, make sure that I understand it properly).

When you say "there are a bunch of parts that start with 'z'", do you mean that users can have
invalid parts in their mappings (that start with 'z') , and that we don't want to replace them
(which is something I asked for in a previous review comment)?  IIUC there are never valid
LCSCs that start with 'z' (e.g. every part in the current catalog starts with C) so this is just
a matter of not surprising the user by replacing an invalid value with something from the part
preference database (and potentially replacing a noisy failure at order time with a silent failure
by ordering the wrong part).

I think this is the right call, and maybe just deserves some comment/docstring to clear it up
(why this particular text matcher  is more permissive than other part matchers).  If this is the
intention, this should be good to merge as is.

As a nitpick/followup, though, I do think that we should make sure we apply this check at
write time to the part preference database - so we don't accidentally round-trip one of these
bad identifiers.  (e.g. we shouldn't overwrite something that the user has entered as a part
number, but we can be opinionated about whether or not we propagate that in the part
preference db)

The CSV importer path can also write these bad values; I don't know if anyone actually uses
this or not but we probably could also validate the parts since they take a direct write path.
Or we could just merge the write path into the batch version and do the validation there.
(Or just ignore it, or kill this feature? Is it used?)

A smaller nitpick is that by newly scanning other fields that contain LCSC/JLC (has_lcsc_assignment_text),
we might not autofill components if the user has added something like "JLC Rotation: 90" or something. 
That's probably minor but might be confusing.  (Maybe just log feedback if this happens?
"Skipped R12 due to JLC-shaped field {field} with value {value}")

PR #810, opened 2026-09-06
Derive enrichment-target classifications from ComponentProductType

z2amiller:
## Summary

Follow-up to the nitpick I left on #804. After that enum refactor, `Store.get_assembly_enrichment_targets()` was the last place still spelling out the numeric classifications:

```python
"OR component_product_type NOT IN (0, 1, 2)"
```

That is the spot most likely to drift — a newly observed classification would need an enum member *and* a hand-edit here, and nothing would fail loudly if you forgot the second one.

- Build the list from `ComponentProductType` instead of repeating its values.
- Bind the values as parameters rather than interpolating them, using the same `?`-placeholder pattern the reference filter directly below already uses. The classification placeholders precede the reference ones in the query text, so their params lead `params`.

Generated query and binds are unchanged in effect:

```
... OR component_product_type NOT IN (?,?,?)) AND reference IN (?,?) ORDER BY lcsc, reference
params: [0, 1, 2, 'R1', 'R2']
```

## Validation

- 455 tests pass; `ruff check` and `ruff format --check` clean on both touched files.
- `test_get_assembly_enrichment_targets_uses_or_logic` already pinned this behavior (classification `0` and `2` count as enriched, `3` and `"bad"` stay targets). It had no row for classification `1`, so I added one — every enum member is now covered.

No behavior change, no schema change.

### DIFF
--- a/common/test_store_pad_filter.py
+++ b/common/test_store_pad_filter.py
@@ -241,7 +241,11 @@ def test_assembly_metadata_follows_lcsc_lifecycle(tmp_path):
 
 
 def test_get_assembly_enrichment_targets_uses_or_logic(tmp_path):
-    """Rows missing any required enrichment field should be selected."""
+    """Rows missing any required enrichment field should be selected.
+
+    Every ComponentProductType member (0, 1, 2) counts as enriched; only NULL
+    or an unrecognized classification keeps a row in the target set.
+    """
     s = _store_obj()
     s.logger = logging.getLogger(__name__)
     s.dbfile = str(tmp_path / "project.db")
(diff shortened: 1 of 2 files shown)


z2amiller:
Closing in favor of #817, which lands the same fix in a tighter diff (+4/-1). Thanks @chmorgan.

I checked out #817 and confirmed it fully covers what this PR did:

- Same generated SQL — `... OR component_product_type NOT IN (0,1,2)` — so no behavior change, same as here.
- It uses `str(member.value)` rather than `str(member)`, which is the version-safe choice and worth calling out: this project supports Python 3.10 (`requires-python = ">=3.10"`), and `IntEnum.__str__` only started returning the bare number in 3.11. On 3.10 `str(member)` would have emitted `NOT IN (ComponentProductType.ECONOMIC_AND_STANDARD, ...)` and broken the query. `.value` is a plain `int` on every version.
- It adds the same missing test row for classification `1`, so `test_get_assembly_enrichment_targets_uses_or_logic` now covers every enum member.

The only real difference is that this PR bound the values as parameters instead of interpolating them. That was to match the `?`-placeholder pattern the reference filter below uses, but the values here are compile-time enum constants rather than caller input, so interpolating them is fine and keeps the query text a self-contained literal. #817's approach is the simpler one and matches what I originally suggested on #804.

474 tests, `ruff check`, `ruff format --check` and `git diff --check` all pass on #817 for me too.

PR #812, opened 2026-09-06
Bind clean_database reference parameters

z2amiller:
`clean_database` built its `DELETE` by f-string-interpolating board reference designators into single-quoted SQL literals. `get_valid_footprints` uses an unanchored `re.match(r"[\w\d-]+", ...)`, so a reference only has to *start* with a word character — a quote passes straight through.

**Scope note:** this originally also parameterized the three correction-table queries. Those are dropped — #816 parameterizes all three and adds finite-value validation for offsets, so it owns that code. Keeping both would have conflicted in `library.py` and on `tests/test_library_corrections.py`. This PR is now `store.py` only and does not touch `library.py`.

### The defect

An apostrophe in a designator breaks the statement. `clean_database` runs from `Store.__init__`, so it raises before the plugin window can open, with nothing in the UI to explain why. Reproduced on `main`:

```
sqlite3.OperationalError: near "A": syntax error
```

Short of a syntax error, a crafted designator silently widens the keep-list. `sqlite3.execute` refuses to chain statements, so this is wrong results rather than arbitrary SQL: a reference like `R1', 'C9` smuggles an extra entry into the `NOT IN` list, sparing an unrelated stale part while deleting the live one.

### The fix

Build only `?`-placeholders into the f-string and bind the references, matching the existing approach in `get_assembly_enrichment_targets`.

**Empty board unchanged.** `",".join` produces an empty list either way, so the statement is byte-identical to before, and SQLite accepts `NOT IN ()` as an extension. There is now a regression test pinning that, per review feedback.

**Variable limit.** Bound parameters trade `SQLITE_MAX_SQL_LENGTH` for `SQLITE_LIMIT_VARIABLE_NUMBER`, one per footprint. That is 32766 on SQLite ≥ 3.32; KiCad's bundled Python 3.9.13 ships SQLite 3.37.2, so it would take a 32,767-footprint board to reach. Not chunking, but recording the reasoning.

### Tests

Four tests on real SQLite, no mocking of the code under test. Two fail on `main` (quoted designator, keep-list widening); the quoted-removal and empty-board cases are guards that pass on both.

```
python -m pytest -q          486 passed
ruff check .                 clean on the changed files
```

`ruff format --check` is clean on both changed files. Rebased onto current `main`.

Not covered: no KiCad-in-the-loop run, so whether KiCad's own reference-field validation permits an apostrophe in a designator is untested — only that the plugin's filter lets one through.

### DIFF
--- a/store.py
+++ b/store.py
@@ -444,10 +444,14 @@ class Store:
 
     def clean_database(self):
         """Delete all parts from the database that are no longer present on the board."""
-        refs = [f"'{fp.GetReference()}'" for fp in get_valid_footprints(self.board)]
+        refs = [fp.GetReference() for fp in get_valid_footprints(self.board)]
         with contextlib.closing(sqlite3.connect(self.dbfile)) as con, con as cur:
+            # The f-string only injects ?-placeholders; the reference text
+            # itself is bound, so quotes in a designator stay literal.
+            placeholders = ",".join("?" for _ in refs)
             cur.execute(
-                f"DELETE FROM part_info WHERE reference NOT IN ({','.join(refs)})"
+                f"DELETE FROM part_info WHERE reference NOT IN ({placeholders})",
+                refs,
             )
             cur.commit()
 
--- a/tests/test_store_clean_database.py
+++ b/tests/test_store_clean_database.py
@@ -1,0 +1,114 @@
+"""Keep board reference text out of the clean_database DELETE statement."""
+
+import logging
+
+import pytest
+
+from .wx_harness import load, package_stubs, wx_stubs
+
+_PACKAGE = "_store_clean_database_tests"
+
+# get_valid_footprints only requires a reference to start with a word
+# character, so quotes reach clean_database untouched.
+QUOTED_REFERENCE = "U1'A"
+
+# sqlite3.execute refuses to chain statements, so the reachable damage is a
+# reference that smuggles an extra entry into the NOT IN keep-list.
+SMUGGLING_REFERENCE = "R1', 'C9"
+
+
+class _Footprint:
+    """Expose the single board accessor clean_database relies on."""
+
+    def __init__(self, reference):
+        self.reference = reference
+
+    def GetReference(self):
+        """Return the board reference designator."""
+        return self.reference
+
+
+class _Board:
+    """Return a fixed set of footprints for reference collection."""
+
+    def __init__(self, references):
+        self.footprints = [_Footprint(reference) for reference in references]
+
+    def GetFootprints(self):
+        """Return every footprint currently on the board."""
+        return self.footprints
+
+
+@pytest.fixture
+def make_store(tmp_path):
+    """Build a Store on real SQLite without the constructor's board sync."""
+    stubs = {**package_stubs(_PACKAGE), **wx_stubs()}
+    store_module = load(_PACKAGE, "store", stubs)
+
+    def _make(board_references, stored_references):
+        store = store_module.Store.__new__(store_module.Store)
+        store.logger = logging.getLogger(_PACKAGE)
+        store.dbfile = str(tmp_path / "project.db")
+        store.board = _Board(board_references)
+        # Constructor defaults that read_all interpolates into its ORDER BY.
+        store.order_by = "reference"
+        store.order_dir = "ASC"
+        store.create_db()
+        for reference in stored_references:
+            store.create_part(
+                {
+                    "reference": reference,
+                    "value": "10k",
+                    "footprint": "R_0603",
+                    "lcsc": "",
+                    "exclude_from_bom": 0,
+                    "exclude_from_pos": 0,
+                }
+            )
+        return store
+
+    return _make
+
+
+def _references(store):
+    """Return the references still recorded in the project database."""
+    return sorted(part["reference"] for part in store.read_all())
+
+
+def test_quoted_reference_survives_cleanup(make_store):
+    """Keep a part whose reference contains a quote and is still on the board."""
+    store = make_store(
+        ["R1", QUOTED_REFERENCE],
+        ["R1", QUOTED_REFERENCE, "C9"],
+    )
+
+    store.clean_database()
+
+    assert _references(store) == sorted(["R1", QUOTED_REFERENCE])
+
+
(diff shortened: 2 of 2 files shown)


library.py
```diff
@@ -383,7 +387,9 @@ def update_correction_data(self, regex, rotation, offset):
             con as cur,
         ):
             cur.execute(
-                f"UPDATE correction SET rotation = '{rotation}', offset_x = '{offset[0]}', offset_y = '{offset[1]}' WHERE regex = '{regex}'"
+                "UPDATE correction SET rotation = ?, offset_x = ?, offset_y = ? "
+                "WHERE regex = ?",
+                (rotation, offset[0], offset[1], regex),
```
chmorgan: **[P2] Preserve readable correction values when binding numeric fields**

Binding native numeric values changes behavior for inputs the correction manager currently accepts. Entering `nan` in an existing correction's offset field produces `float("nan")`, which SQLite binds as NULL. The update commits, then `get_all_correction_data()` raises `TypeError` on `float(None)`. Subsequent correction reads also fail.

I reproduced this through the real `CorrectionManagerDialog.save_correction()` method with stubbed controls and real SQLite on both revisions:
- Base: saves successfully, stores `("nan", "text")` for `offset_x, typeof(offset_x)`, and subsequent reads succeed.
- PR: stores `(None, "null")`, then list refresh and subsequent reads raise `TypeError: float() argument must be a string or a real number, not 'NoneType'`.

Please preserve the previous string representation when binding the numeric fields, or reject nonfinite values before committing, and add a regression test. Parameterizing the query still fixes the quoting problem either way.

The trigger is unusual, and insertion already has the same weakness, but this change extends that persistent failure to editing existing corrections.

chmorgan (changes requested):
Reviewed commit `4d5b84706a5804b194682c1cbdb7fea94e221fe1`, including the context from #809.

The parameterized queries fix the quoted-input failures and unintended row selection. I am requesting one small change before merge: the switch to native numeric bindings introduces a persistent correction-read failure when an existing offset is edited to `nan`. Details and the reproduction are in the inline comment on `library.py`.

Optional test suggestion: add an empty-board cleanup regression. Seed stored references, supply no valid board footprints, call `clean_database()`, and assert that no references remain. I verified that the current implementation handles this correctly; the test would protect that special case.

Validation on the exact PR commit, using the repository's existing `myenv`:
- All 462 suite tests passed across runs. Five initially encountered sandbox restrictions on temporary localhost HTTP servers; rerunning the file-manager tests with the required access passed.
- Six of the seven new tests fail on the base revision; all seven pass on the PR.
- Ruff lint and formatting checks pass for all four changed files.

The dialog-method reproduction uses stubbed GUI controls and real application methods/SQLite. Live KiCad UI validation remains untested.

z2amiller:
Thanks — reproduced the nan case exactly as you describe. Native `float("nan")` binds as NULL (`typeof` → `'null'`) where the f-string stored TEXT `'nan'` that `float()` round-trips. Worth noting it's offsets only: `save_correction` does `int(self.to_float(...))` for rotation, so `int(float("nan"))` raises `ValueError` before reaching the database. `inf` binds fine as REAL, so NaN is the sole divergence.

Rather than preserve the string representation, I've dropped the correction changes from this PR entirely. Looking at #816, it already deletes the same three f-string statements and replaces them with bound parameters, and it fixes nan properly at the validation layer — `_offset` rejects nonfinite via `math.isfinite`, reached from `save_correction` through `validate_correction`. Keeping my version would have duplicated that and conflicted in two places: a content conflict in `library.py` and an add/add on `tests/test_library_corrections.py`, which we both added.

So #816 owns the correction queries. This PR is now `store.py` only and doesn't touch `library.py` at all. I confirmed the two branches merge onto current `main` with zero conflicts and 966 tests pass together.

Added your empty-board regression test — you were right that it passes on both revisions, so it's a guard rather than a regression, and the statement really is byte-identical either way since `",".join` yields an empty list before and after.

Also rebased onto current `main`, which had moved on. One thing worth flagging separately: #809 merges cleanly into current `main` but 27 of its own tests then error with `ImportError: cannot import name 'ComponentProductType'` — #817 added that import to `store.py`, and #809's tests are the first to load the real `store.py` through `mainwindow_stubs`, whose `bom_estimation.assembly_mode` stub doesn't provide it. Adding it to that stub takes it to 527 passed. Neither CI run can catch it since they never run together.

chmorgan (approved):
Re-reviewed `27634aa8d5103d10193d35e0e138ea089e10d6cb`. No blocking findings in the remaining diff; approving and clearing my earlier request for changes.

The previous NaN finding no longer applies to #812: `library.py` is byte-identical to the current base, and this PR now changes only `Store.clean_database` and its tests. Correction validation is left to #816; this approval covers the narrowed #812 patch.

The empty-board guard requested in the previous review is now included and passes. I also checked the remaining patch independently of the earlier review.

Evidence from the exact PR snapshot, using the repository's existing `myenv`:

- **Literal reference handling:** all four cleanup tests pass. On the base revision, the quoted-reference test fails with the expected SQL syntax error and the keep-list test incorrectly retains `C9`; the quoted-removal and empty-board guards pass on both.
- **Initialization and persisted data:** additional checks through real `Store.__init__`, using stateful board/footprint doubles and real SQLite, passed when opening a database created by the base revision. Live quoted and SQL-like references survive, stale rows are removed, and saved LCSC assignments, stock, exclusion flags, and generation metadata are retained.
- **Reopening and later board changes:** reopening preserves those results; removing quoted footprints deletes their stored rows on the next initialization. Boards with only filtered references, and empty boards, clear part rows while preserving project metadata.
- **Failure path:** a trigger-aborted DELETE leaves all part rows intact rather than committing a partial cleanup.
- **Checks:** 486 tests passed; Ruff lint and formatting checks passed for both changed files.

Native KiCad reference-field/UI behavior was not exercised: `pcbnew` is unavailable in the required environment. The constructor checks above use board/footprint doubles, not a live KiCad session. The documented per-footprint SQLite parameter-limit tradeoff remains.

Bouni:
@z2amiller and @chmorgan I'm totally fine if you guy review each others PRs and merge them. I have so little spare time at the moment that I'm too much of a blocker for this project. Feel free to bring it forward!

Bouni (approved):

PR #816, opened 2026-09-06
Validate correction imports and recover invalid stored corrections

### LINKED ISSUE #531: Crash after trying to import rotations file
**Describe the bug**
I can't use the plugin anymore. When I created a new project, I tried to import the rotations file from another project. After that I see this:
image

**To Reproduce**
1. Go to the rotations option
2. Add an exported CSV file with corrected rotations
3. Leave KiKad and try to open it again.

**Expected behavior**
I expected to be able to load the plugin after that change.

# KiCad Version

```
Application: KiCad PCB Editor x64 on x64

Version: 8.0.4, release build

Libraries:
	wxWidgets 3.2.5
	FreeType 2.12.1
	HarfBuzz 8.3.0
	FontConfig 2.14.2
	libcurl/8.5.0-DEV Schannel zlib/1.3

Platform: Windows 11 (build 22631), 64-bit edition, 64 bit, Little endian, wxMSW

Build Info:
	Date: Jul 17 2024 03:26:47
	wxWidgets: 3.2.5 (wchar_t,wx containers)
	Boost: 1.83.0
	OCC: 7.8.1
	Curl: 8.5.0-DEV
	ngspice: 42
	Compiler: Visual C++ 1939 without C++ ABI

Build settings:

```

chmorgan:
Importing a malformed correction CSV could persist values such as `47u`, then crash the plugin when corrections reloaded or the project reopened. Validate complete input before atomic writes, keep older invalid records repairable, and check a complete correction snapshot and its placements before fabrication starts.

Fixes #531.

### Changes

- Make `Correction` an immutable validated value with shared storage, CSV, editor, and display adapters. Retain both compiled regex forms and share suffix-first reference/value/footprint matching between the list and CPL. Each included footprint is matched once.
- Preserve original invalid database records for repair. Guard manager edits, deletes, and replacements against concurrent record changes inside the write transaction; retain entered text when an operation must be retried.
- Keep legacy inspection out of ordinary reads. Preserve existing destination patterns, then rotations-archive priority over parts archives and initial defaults. Record completed examination, retain known pending transfers across reopening, and report unreadable unknown archives as nonblocking warnings.
- Make optional initial-download scheduling failures warnings. Preserve source archives, atomic transfers and completion records, resumable initial downloads, and completed migrations that do not replay over user deletions. Correct UNC URI handling at both database-open boundaries.
- Prepare all placement rows before generation side effects, reporting unrepresentable coordinates without overwriting previous output. Emit the prepared rows without rereading the board or matching again.

### Acceptance evidence

| Requirement | Evidence |
| --- | --- |
| Malformed imports cannot poison storage | Real SQLite and production import/manager workflows cover `47u`, numeric and regex boundaries, supported historical/current CSV formats, physical line diagnostics, duplicate policies, full-batch rejection, rollback, repair, and reopening. |
| Repair affects the selected and approved records | Constructor/handler regressions cover deleted and reused row IDs, same-row edits, raw SQLite type changes, and changed/added/removed replacement conflicts. Dirty inputs retain their original expected record. Failed actions refresh displayed rows; explicit reselection accepts current data. |
| Display and CPL use the same usable values | Immutable construction/replacement and serializer tests remain. Large-rule tests purge Python's regex cache and require no compilation during matching. Actual CPL/list workflows cover suffix-first ranking, target priority, ties, zeros, no match, both layers, offsets, and one selection per included footprint. |
| Placement failures precede output changes | Real exporters cover extreme finite offsets, final coordinate boundaries, and valid corrections that cancel a large origin-relative intermediate. Existing CPL bytes remain intact and absent files remain absent on failure. Production generation-handler tests stop before fill/hooks/Gerber/drill/archive steps. |
| One generation stays coherent; the next sees fresh data | The real generation/CPL workflow changes storage and board positions after preflight, verifies the exact prepared output and match count, blocks the next invalid generation before side effects, then verifies recovery after repair. |
| Migration does not disable unrelated healthy corrections | Tests require source-free snapshots and read-only settled startup. Missing/empty/no-table sources are terminal. Unknown archives warn; known incomplete transfers remain blocking across reopening and source loss. Optional thread-start failures preserve usable corrections through startup, manager retry, and committed scope changes. |
| Migration preserves data and precedence | Real transactions cover source priority, established repairs, malformed within-source records, rollback and completion failures, interrupted seeding, concurrent download ownership, changed library selection, and canonical path aliases. NULL/BLOB metadata keys cannot crash recovery or discard pending diagnostics. |
| Path and framework assumptions have explicit evidence | Portable SQLite URI tests exercise UNC/localhost identity, escaping, read-only/existing-only modes, missing files, and deletion races. Stateful wx doubles exercise actual manager constructors and handlers; authoritative wx/KiCad source supports selection and coordinate contracts. |

### Simplification and review

The review revision before the #812 rebase, compared with `2a54dec`, has **12 fewer production Python lines**. Regression tests add **212 lines** and documentation adds three; the full tracked revision is therefore **203 lines larger**.

The reduction removes redundant manager selection/row/state fields, manual multi-selection handling, an unused private matcher wrapper, duplicate constructor/rotation logic, an unused migration argument, and repeated metadata reads and completion/scheduler handling. Repeated test setup was consolidated; a weaker mocked generation test was replaced by the stronger real-output workflow. Distinct recovery cases were retained.

Review included a parent acceptance trace, an independent requirements-first audit, and an independent final patch review. Findings about source aliases, malformed metadata keys, final versus intermediate coordinate bounds, selection events, and in-flight download ownership were reconciled with regression coverage. Review revisions are folded into the original implementation commit.

### Validation and limits

Rebased onto upstream `main` at `0ad15dc` after #812 landed. An independent review and `git range-diff` confirm the original correction patch is unchanged. The full suite includes the four new database-cleanup tests from #812.

- Existing repository `myenv`: **1,282 tests passed**, no exclusions. Repository-wide Ruff lint, formatting and Python 3.9 syntax checks for all 19 PR Python files plus both incoming #812 files, annotations on all 115 new/modified revision functions, and whitespace checks passed.
- Configured Ruff, formatting, pyupgrade, and markdownlint commit hooks passed. Final committed files exactly match the reviewed and tested bytes; local `AGENTS.md` and `myenv` remain excluded.
- Tests use Python 3.12, real SQLite and CSV/CPL files, production constructors/handlers, and simulated wx/pcbnew objects. The installed KiCad Python conversion helper was also exercised without loading native KiCad.
- Native wx creation was attempted in `myenv` but unavailable because that interpreter lacks Framework/display access. Native KiCad, Windows/SMB, and Python 3.9 runtime execution are not claimed; no build or alternate environment was used.
- The original issue CSV was unavailable. Older versions also did not record whether absent patterns were deleted, renamed, or never imported; first migration cannot reconstruct that provenance. Completed migrations preserve later deletions, and explicitly restored data can be imported through supported CSV formats.

### DIFF
--- a/CORRECTIONS.md
+++ b/CORRECTIONS.md
@@ -1,0 +1,141 @@
+# Importing and repairing corrections
+
+The Corrections Manager stores rotation and position corrections for footprint,
+value, or reference patterns. Corrections can use a global database or a database
+in the current project's `jlcpcb` directory.
+
+## CSV formats
+
+The current exported format is:
+
+```csv
+Pattern,Rotation,Offset X,Offset Y
+^SOT-23$,-90,0,0
+^USB_C_Receptacle$,0,1.44,0
+```
+
+Historical files using `Footprint pattern,Correction` and downloaded files using
+`Footprint pattern,Rotation,Offset X,Offset Y` are also supported. Recognized
+columns may be reordered. The importer rejects unknown or duplicate columns
+instead of guessing their meaning.
+
+Rotation must be a finite whole number of degrees. For example, `90`, `-90`, and
+`90.0` are accepted; `47u` and `90.5` are rejected. Signed values are preserved.
+Offsets must be finite numbers in millimeters. Patterns must be nonempty and
+valid for the plugin's regular-expression matching.
+
+Missing trailing offsets default to zero, including two-field records beneath
+a four-column header. An explicitly empty numeric cell is an error. A missing
+rotation is always an error. UTF-8 files, including files with a byte-order mark,
+Windows line endings, and ordinary CSV quoting are supported.
+
+A file containing only a recognized header imports no records. Blank lines are
+ignored; an entirely empty file or a row containing only delimiters is rejected.
+
+## Import behavior
+
+The complete file is validated before any corrections are changed. If one row
+is invalid, the entire import is rejected and the existing database stays
+unchanged. The error identifies the source, line, field, and invalid value.
+
+A successful user import updates existing patterns and adds new ones. If the
+same pattern appears more than once, its last occurrence wins. Every row must
+still be valid. Downloading corrections preserves existing entries and adds
+missing patterns only.
+
+Database writes are transactional. A write failure rolls back earlier inserts
+and updates from that operation. Correct the reported problem and import again.
+
+## Repairing existing invalid data
+
+Earlier plugin versions could store invalid numeric text from a CSV. Issue #531
+reported a rotation of `47u`, which then prevented the plugin from reopening.
+
+The manager shows invalid records with their original values and errors. Select
+one to repair or delete it. If another window changes the selected record or a
+record approved for replacement, the operation stops and retains your entered
+text. Reselect the current record before retrying. Repair affects only the
+selected record, even when several records have the same pattern.
+
+The main window remains accessible while correction errors are unresolved.
+Fabrication and correction CSV export require a valid active database; invalid
+corrections are never omitted or replaced with zero. Fabrication also checks
+transformed placement coordinates before generating files. An unrepresentable
+position identifies the footprint and correction to repair, preserving existing
+output files.
+
+An empty correction database is valid. A part with no matching rule uses its
+uncorrected placement. Unreadable storage or unresolved records are shown as
+`Unresolved`, rather than being treated as an empty set of corrections.
+
+## Matching and displayed corrections
+
+The parts table and placement output use the same correction selection rules.
+Reference matches take priority over value matches, which take priority over
+footprint matches. For each of those fields, patterns matching the end of the
+text are tried before ordinary substring matches. Ties retain database order.
+For example, `SOT-23-3` wins over `SOT-23` for a footprint named `SOT-23-3`.
+
+A matched correction of zero degrees and zero offset still takes precedence
+over lower-priority rules. The table labels the match source as `(ref)`, `(val)`,
+or `(fpt)`.
+
+## Legacy data and database scope
+
+Existing patterns in the current correction database take precedence over
+legacy rotation archives. This preserves corrections already repaired or
+customized after an older migration, including their position offsets. A stale
+archive cannot add a conflicting copy of an established pattern.
+
+For patterns absent from the destination, the dedicated `rotations.db` archive
+takes precedence over the selected parts database's old rotation table. A
+lower-priority archive cannot add a conflicting copy. Original values are
+preserved, including invalid values and contradictions within the winning
+archive that need repair. The source archives are retained, and imported rows
+and their completion records commit together.
+
+Migration is checked at startup, when switching to global corrections, and when
+opening Corrections Manager for global corrections. Ordinary list refreshes,
+exports, and fabrication checks do not reopen legacy archives. Missing sources,
+empty rotation tables, and files without a rotation table are recorded as
+examined; they do not block healthy corrections or need repeated examination.
+
+If an archive cannot be read and its corrections are unknown, the manager shows
+a warning identifying the file. Healthy active corrections remain usable.
+Restore access to the archive and reopen Corrections Manager to retry. Automatic
+imports from lower-priority sources and the initial default download wait for
+this retry so they cannot take precedence over recovered custom corrections.
+
+If corrections were identified but could not be transferred, the incomplete
+transfer remains blocking across reopening. The manager identifies the source
+and failure; resolve that failure and reopen the manager to retry. An archive
+disappearing afterward does not make a known incomplete transfer successful.
+
+If a custom database schema would change a legacy value during transfer, the
+whole transfer is rejected and the archives remain intact. Correct the reported
+storage incompatibility before retrying.
+
+Older versions did not record whether a missing pattern was never imported,
+deleted, or renamed. If an old archive remains, a missing pattern can be
+restored during its first migration. Completed migrations do not replay that
+archive over later repairs or deletions. Records restored after a source was
+already examined can be imported explicitly using the supported CSV formats.
+An interrupted initial default download resumes after recovery. Failure to start
+that optional download warns without disabling healthy corrections. Reopening
+does not refill corrections that were deliberately deleted.
+
+Automatic legacy CSV imports archive the source only after the import commits.
+An archive failure is reported separately from a successful database import.
+Rejected CSV input leaves the current stored corrections usable. Warnings about
+unreadable archives are separate from errors in active corrections and known
+incomplete transfers.
+
+Global-to-local copying completes before the active database changes. A failed
+switch preserves the active scope and its data. Errors in an inactive database
+do not block a healthy active database.
+
+Switching back to global corrections discards the project's local corrections
+after an explicit confirmation and a successful check of the global database.
+Invalid local values do not prevent this choice. If global corrections cannot
+be used or the switch fails, the local corrections and active scope remain in
+place. Other project data is preserved.
--- a/README.md
+++ b/README.md
@@ -141,6 +141,9 @@ This tool will automatically export all additional layers with the JLC_ prefix a
 JLCPCB seems to need corrected rotation information. @matthewlai implemented that in his [JLCKicadTools](https://github.com/matthewlai/JLCKicadTools) and I adopted his work in this plugin as well.
 You can download Matthews file from GitHub and manage your own corrections in the Rotation manager.
 
+See [Importing and repairing corrections](CORRECTIONS.md) for supported CSV
+formats, validation rules, and recovery of invalid records from older versions.
+
 ## Icons
 
 This plugin makes use of a lot of icons from the excellent [Material Design Icons](https://materialdesignicons.com/)
--- a/correction_data.py
+++ b/correction_data.py
@@ -1,0 +1,382 @@
+"""Validate correction values and CSV files without GUI or database dependencies."""
+
+from collections.abc import Iterable, Iterator, Sequence
+import csv
+from dataclasses import dataclass, field, replace
+from decimal import Decimal, InvalidOperation
+from io import StringIO
+import math
+import re
+from typing import Literal, Optional
+
+_MIN_ROTATION = -(2**63)
+_MAX_ROTATION = 2**63 - 1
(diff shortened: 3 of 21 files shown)


z2amiller:
Thanks for this — the diagnosis in #531 is right, and `correction_data.py` is the correct
place to fix it. Two structural asks before it lands, plus two bugs I hit while reading.

## 1. Make `Correction` a value type whose absence *is* the validity check

Right now `Correction` is a frozen dataclass with a public constructor and no
`__post_init__`, so it can hold anything:

```python
>>> Correction(pattern="(unclosed", rotation="47u", offset=("x", None))
Correction(pattern='(unclosed', rotation='47u', offset=('x', None))   # accepted
```

Because holding one proves nothing, validity has to be re-established at every use, and the
PR does exactly that. `generate_cpl` re-validates values it just received from
`require_valid()`:

```python
corrections = self.parent.library.read_correction_data().require_valid()
validated = tuple(validate_correction(c.pattern, c.rotation, c.offset) for c in corrections)
self.corrections = [(c.pattern, c.rotation, c.offset) for c in validated]
```

— validate, then immediately discard the type and go back to tuples. `CorrectionSnapshot`
carries `.issues`, `.require_valid()` and `.as_tuples()` for the same reason, and callers
check `if not snapshot.issues` rather than asking whether they have a correction.

The ask: enforce the invariant in the constructor and let `None` mean invalid.

```python
@dataclass(frozen=True)
class Correction:
    pattern: str
    rotation: int
    offset: tuple[float, float]

    def __post_init__(self):
        """Reject anything that is not a usable correction."""
        ...  # raises

    @classmethod
    def parse(cls, pattern, rotation, offset) -> Optional["Correction"]:
        """Return the correction these values name, or None if they name none."""
        try:
            return cls(pattern, rotation, offset)
        except (TypeError, ValueError):
            return None
```

No `.valid()`, no `require_valid()` — a correction in hand is valid. #815 is the same
pattern applied to LCSC part numbers in this repo, if a worked example helps; #807 already
consumes it from the corrections code.

Then let the object render itself for each consumer instead of being taken apart:

| Today | Becomes |
| --- | --- |
| `(c.pattern, c.rotation, *c.offset)` splatted into SQL in 6 places | `c.db_row()` |
| `writer.writerow([c.pattern, c.rotation, *c.offset])` | `c.csv_row()` |
| `f"{rotation}°, {offset[0]}/{offset[1]}"` in `mainwindow` and `_confirm_overwrite` | `str(c)` |
| `str_from_float` in the manager, bare `str(offset[0])` in the list | `c.editor_values()` |
| `snapshot.as_tuples()` | drop it; hand callers `tuple[Correction, ...]` |

That last row is what makes the two matchers collapse into one, which is bug (3) below.

## 2. Where the *reasons* live

The one thing `None` genuinely loses is why a value failed, and this PR needs that — the
repair list is the point of it. I don't think that argues against the type; it argues for
keeping the diagnostics at the two edges that must *explain* rather than *decide*:

- the CSV importer, which reports every bad row in a file with line numbers and rejects the
  whole import — `parse_corrections_csv` keeps raising `CorrectionDataError`;
- the manager's repair list, which shows a stored row beside its errors.

Everywhere else — `fabrication`, `mainwindow`, the matcher, export — takes
`Optional[Correction]` and never sees a `CorrectionIssue`. `StoredCorrection` is already
almost this shape: it holds both `.issues` and `.correction`, so `.correction is None`
becomes the check and `.issues` becomes purely the text the repair column renders.

Worth keeping three absences distinct, since they mean different things to a user:

- no correction matches this part → `"0°, 0.0/0.0"` (a `None` from the matcher)
- this stored row will not parse → repairable, shown in the manager
- the active database is unreadable → `"Unresolved"` (a snapshot-level state, not a
  per-correction one)

## 3. Bug: the parts list and the CPL can disagree about which correction applies

`fabrication._find_correction` tries anchored patterns first, so `SOT-23-3` beats `SOT-23`.
`mainwindow.get_correction` is a plain unanchored `re.search` over reference → value →
footprint. They disagree:

```
patterns: [("SOT-23", 180), ("SOT-23-3", 90)]
part: ref Q1, value MMBT3904, footprint SOT-23-3

  UI "Correction" column shows : 180° (fpt)
  CPL actually applies         :  90°
```

Pre-existing on `main`, not introduced here — but it is the defect a single type removes,
and #805 changes that ranking again. One `find_correction(corrections, value)` over
`tuple[Correction, ...]`, used by both, is the fix.

## 4. Bug: the legacy migration can lock a user out of generating fabrication files

`_migrate_correction_source` inserts every legacy `rotation` row unconditionally — it does
not skip patterns the target already holds. `main` deleted `rotations.db` after migrating,
so any user whose migration previously failed partway still has the file. On upgrade those
rows come back, `_mark_conflicting_corrections` flags the pattern, `require_valid()` raises,
and `read_valid_corrections_for_generation` is now the first step of Generate:

```
rows after upgrade:
   1 ^SOT-23$ -90 (0.0, 0.0)  -> 1 issue(s)   ← the user's own repair
   2 ^SOT-23$ 180 (0, 0)      -> 1 issue(s)   ← re-migrated legacy value
   3 ^USB_C_Receptacle$ 0     -> 0 issue(s)
RESULT: fabrication generation BLOCKED
```

`CORRECTIONS.md` documents conflict preservation as deliberate, so this may be intended —
but the effect is a hard block arriving from a migration the user did not ask for.
`INSERT ... WHERE NOT EXISTS (SELECT 1 FROM correction WHERE regex = ?)` preserves repairs
and never manufactures the conflict.

Smaller, same area: `switch_to_global_correction_database(True)` still drops the local table
without copying it anywhere, but now first requires that data to validate —
`source = source_snapshot.require_valid()`, and `source` is then unused in that branch. So a
user must repair corrections before being allowed to discard them.

## 5. Please use `tests/wx_harness.py` rather than a second copy of it

`correction_test_support.load_correction_modules` rebuilds the synthetic package
(`package_stubs`), the fake wx and its submodules (`wx_stubs`), and the
import/pop/restore dance (`temporary_modules`) by hand.
`test_fabrication_correction_recovery.py` repeats the third one again;
`test_mainwindow_correction_recovery.py` uses both scaffolds at once, so two unrelated fake
`wx` modules coexist in one test.

The copy is also weaker in a way that has already cost something. Its constants are a fixed
17-name table numbered `enumerate(..., 1)`, so they are sequential integers rather than
distinct bits — `wx.OK|wx.ICON_ERROR` is `5|7 = 7`, `wx.YES_NO|wx.NO_DEFAULT` is `10|11 =
11`. `FakeWxModule` mints `1 << n` on demand precisely so style arithmetic cannot alias and
so a missing constant is never a maintenance task. That maintenance already showed up:
`test_corrections_import_export.py:139` carries
`monkeypatch.setattr(modules.wx, "ICON_QUESTION", 1 << 24, raising=False)` for the one
constant the table forgot.

Swapping the hand-rolled block for `package_stubs()` + `wx_stubs(NOT_FOUND=-1, Dialog=...,
PostEvent=..., MessageBox=..., MessageDialog=...)` and deleting that `ICON_QUESTION` line
passes the suite unchanged — about 39 lines become 18.

The harness genuinely is missing what you needed, though, and it is worth adding rather than
working around: `load()` is one-shot, and you need several *real* sibling modules loaded
together with their relative imports intact and left registered. A `load_siblings(package,
names, replacements)` next to `load()` would cover it, and then
`correction_test_support.py` keeps only what is actually about corrections — `make_library`,
`fresh_library`, `seed_raw`, `raw_rows`.

---

For what it's worth: `ruff check .` passes under the repo's pinned 0.14.14 and every file
this PR touches is format-clean; full suite is 954 passed here.

z2amiller:
Thanks for fixing that - a couple more things, sorry, this is a big one:

### Regex-cache cliff — pre-existing in the CPL, new in the parts list

find_correction (correction_data.py:88-98) needs 2N entries in re's 512-slot cache and scans them cyclically. Measured, 200 parts × 3 targets: 60 corrections 0.012 s, 200 → 0.11 s, 300 → 4.1 s. main's CPL matcher had the same shape; the parts list used the bare form (cliff at 512) and now uses match_correction (cliff at 256) on every populate_footprint_list: 79 ms on main, 3.8 s here at 300 rows. The constructor already compiles both forms (:208-209) and discards them; keep them on the object and search with them (17 ms at 300, no cache dependence). Separately, the CPL matches twice per footprint (fabrication.py:185, :241); CorrectionMatch carries both, so once would do.

The real issue is that it's easy to reconstruct new regexp patterns in Python and not store the compiled forms because Python keeps a small-ish cache under the hood, but once you exceed that cache, the performance trainwrecks.  

 ### Windows UNC (network) paths can break. 

_correction_transaction uses .as_uri() for the path, where other database helpers don't use this, they just use the raw path -- apparently this breaks on Windows remote filesystems ("file://nas/pcb/proj/jlcpcb/project.db'" can't be opened by sqlite3)
 
### Legacy-source probing runs on every read and blocks fabrication on unrelated files — new
_pending_migration_issues (library.py:1320-1351) opens rotations.db and parts.db inside every read_correction_data(). The completion marker is only written for a source that had rows (:1377-1378, :1429-1431), so a source with no table — every current parts.db — is probed on every refresh, save, export and generate. Its errors land in the same issues list (:842), so:

parts.db mid-extraction during a library update: valid corrections.db reads UNAVAILABLE, Generate blocked until extraction finishes.
a rotations.db with no rotation table (zero bytes suffices): required=True raises (:1306-1309), no marker, permanent block. The banner says "Select a row to repair or delete it" (corrections.py:473) but no row action, re-import or scope switch clears it; only deleting the file does. main skipped this silently.
established (:1362-1367) is never updated, so a pattern in both archives is inserted twice — differing values put a fresh upgrade into NEEDS_REPAIR while the log reports success. (main inserted duplicates too; the block is new.)
check_library (:203) also now takes BEGIN IMMEDIATE on corrections.db at every startup to discover nothing is pending, racing the remote-fetch thread started three lines later.

Fix: write the marker for every source examined, treat a missing or unreadable archive as nothing to migrate, add inserted patterns to established, and name the file in the message. Simpler still: rename the archive to rotations.db.migrated on success, which removes the marker table and the per-read probe entirely.

z2amiller:
Thanks for the fixes, everything checks out!

PR #818, opened 2026-09-06
Remember BOM column widths and part selector sizing

### LINKED ISSUE #617: Save sizing of part search window
When searching for parts in "Assign LCSC number" the window size and the width of columns always revert to a standard value. It would be really useful to save these for the next search query, it is kind of annoying to always resize everythings for every part.

### LINKED ISSUE #651: Column widths dont hold a user width
after resizing columns to see all data inc BOM/POS in first window or stock in select parts window apon next entry to the same screen the columns default back to a 'default' width and not what they were set at causing info to fall out of view constantly. 

# KiCad Version
as per #649

chmorgan:
Persist BOM/POS and part-selector column widths and the selector’s latest normal window size through the existing settings file, including when the parent window closes.

Stable column keys survive schema changes; an unpersisted spacer absorbs stretching. Restore widths after layout and bound them to the table client width. Bound selector geometry to the parent display’s work area, using a 1100×600 DIP minimum capped for smaller displays. Preserve normal geometry during maximize/restore and fullscreen. Save complete JSON through atomic replacement, and finish window cleanup when capturing or saving fails.

Fixes #651
Fixes #617

| Requirement | Evidence |
| --- | --- |
| Widths and normal size survive reopening/search | Constructor/event-handler regressions and native main/selector workflows cover BOM/POS, Stock, JSON reopening, retarget/search, and combined parent close. |
| Restored geometry is usable and corrected values persist | Native `[1,1]` becomes `1100×600`; oversized values and `10**100` fit the work area. Three-monitor reopening, including Retina/non-Retina transitions, keeps the window inside the target display. Controls have nonzero bounds at the minimum. Missing displays, smaller work areas and fractional/legacy DIP conversion have regressions. |
| Columns retain identity and bounded widths | Tests cover schema insertion/removal/reordering, malformed and obsolete keys, fixed-column/spacer exclusions, and layout timing. Native extreme widths cap to the actual table client width; chosen Price 137 survives enlarge/shrink/save/reopen. |
| Window-state transitions retain normal size | Six native normal/maximize/restore sequences include animation-time closes; four fullscreen close/restore/resize sequences and fullscreen parent close pass. Regressions also cover iconization, all three capture entry points, and destroyed callbacks. |
| Failed saves preserve settings and complete cleanup | Real-file tests cover partial writes, TypeError, circular-data ValueError, replacement failure, byte-preservation/reload, and temporary cleanup. Close regressions cover capture/serialization failures, ownership clearing, modal/modeless destruction, and incomplete initialization. Combined shutdown saves once. |

Tests remain in a separate commit. Validation uses the existing `myenv`: **1,372 tests pass**, including **85 layout cases** and **5 real-file persistence cases**; Ruff lint/format, whitespace, and Python 3.9 syntax checks pass. Reported failures were reproduced before their fixes. Requirement tracing, independent review starting from the issues, and final patch/commit review are complete.

Rebased onto `main` at `3b3905b` after #816 and #823 landed. Upstream correction recovery and generation validation remain intact. Native main/selector constructors, search, and combined close were rechecked; showing/hiding #816’s correction-status row and resizing preserve restored widths. All GitHub CI checks pass at rebased head `d107fdb`.

Native evidence uses standalone **wxPython 4.3.1 / wxWidgets 3.3.3 Cocoa**, with external data/settings services substituted. Embedded KiCad, reported wx 3.2.8, Windows/GTK—including the possible GTK event-ordering race—and native iconization remain unverified. Screenshot capture was unavailable; label rendering was not inspected. Small displays retain the existing fixed filter layout. Concurrent-window merging and recovery of already-corrupted JSON remain outside scope.

### DIFF
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -1,5 +1,7 @@
 """Contains the main window of the plugin."""
 
+from __future__ import annotations
+
 # pyright: reportMissingImports=false, reportMissingModuleSource=false
 # ruff: noqa: I001
 
@@ -7,12 +9,13 @@ from collections.abc import Sequence
 from contextlib import contextmanager, suppress
 from datetime import datetime as dt
 from threading import Thread
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Optional
 import json
 import logging
 import os
 import re
 import sys
+import tempfile
 import time
 
 import pcbnew as kicad_pcbnew
@@ -79,6 +82,13 @@ from .schematicexport import SchematicExport
 from .settings import SettingsDialog
 from .store import Store
 from .why_standard_dialog import WhyStandardDialog
+from .window_layout import get_column_widths, restore_column_widths
+
+FOOTPRINT_COLUMN_KEYS = {
+    index: key
+    for key, index in PartListDataModel.columns.items()
+    if key not in {"TRAILING_SPACER_COL", "STANDARD_ONLY_COL"}
+}
 
 if TYPE_CHECKING:
     from .library import CorrectionSnapshot
@@ -124,7 +134,9 @@ class JLCPCBTools(wx.Dialog):
     """JLCPCBTools main dialog."""
 
     def __init__(
-        self, parent: Any, kicad_provider: KicadProvider = KicadProvider()
+        self,
+        parent: Optional[wx.Window],
+        kicad_provider: KicadProvider = KicadProvider(),
     ) -> None:
         while not wx.GetApp():
             time.sleep(1)
@@ -152,8 +164,7 @@ class JLCPCBTools(wx.Dialog):
         self.load_settings()
         # Normalize and write-back BOM-estimator settings into the in-memory
         # dict so subsequent reads see canonical values. The on-disk JSON is
-        # not rewritten here — the next save_settings() call (e.g. when the
-        # user changes a setting via the UI) persists these defaults.
+        # not rewritten here; the next settings change or window close saves it.
         general_settings = self.settings.setdefault("general", {})
         raw_board_count = general_settings.get("bom_estimator_boards", 5)
         try:
@@ -648,6 +659,12 @@ class JLCPCBTools(wx.Dialog):
         self.bom_widget.set_visible(self.bom_estimator_show)
         self.Layout()
         self.Centre(wx.BOTH)
+        restore_column_widths(
+            self.footprint_list,
+            self.settings.get("mainwindow", {}).get("column_widths", {}),
+            FOOTPRINT_COLUMN_KEYS,
+        )
+        self._layout_ready = True
 
         # ---------------------------------------------------------------------
         # ------------------------ Custom Events ------------------------------
@@ -770,19 +787,38 @@ class JLCPCBTools(wx.Dialog):
             return {}
         return self.library.get_part_details(lcsc)
 
-    def quit_dialog(self, *_):
+    def quit_dialog(self, *_: object) -> None:
         """Destroy dialog on close."""
-        self.logger.info("quit_dialog()")
-        if self._why_standard_dialog is not None:
-            self._why_standard_dialog.Close()
-        root = logging.getLogger()
-        with suppress(AttributeError):
-            root.removeHandler(self.logging_handler1)
-        with suppress(AttributeError):
-            root.removeHandler(self.logging_handler2)
-
-        self.Destroy()
-        self.EndModal(0)
+        logger = logging.getLogger(__name__)
+        logger.info("quit_dialog()")
+        layout_ready = getattr(self, "_layout_ready", False)
+        selector = getattr(self, "_part_selector", None)
+        try:
+            if layout_ready:
+                self.settings.setdefault("mainwindow", {})["column_widths"] = (
+                    get_column_widths(self.footprint_list, FOOTPRINT_COLUMN_KEYS)
+                )
+                if not selector:
+                    self.save_settings()
+        except OSError:
+            logger.exception("Unable to save window layout")
+        finally:
+            try:
+                if selector:
+                    # Its close handler saves both windows' updated settings once.
+                    selector.Close()
+            finally:
+                why_standard_dialog = getattr(self, "_why_standard_dialog", None)
+                if why_standard_dialog:
+                    why_standard_dialog.Close()
+                root = logging.getLogger()
+                with suppress(AttributeError):
+                    root.removeHandler(self.logging_handler1)
+                with suppress(AttributeError):
+                    root.removeHandler(self.logging_handler2)
+                if self.IsModal():
+                    self.EndModal(0)
+                self.Destroy()
 
     def init_library(self):
         """Initialize the parts library."""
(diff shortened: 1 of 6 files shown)


z2amiller (commented):
Reviewed against `main` locally: full suite passes (505 passed, 31 of them new), `ruff check` and `ruff format --check` clean on the changed files, CI green.

## Looks good

- **Packaging** — `PCM/create_pcm_archive.sh` copies `./*.py` by glob, so the new top-level `window_layout.py` ships without a manifest edit.
- **Python 3.9** — `dict[str, int]` is PEP 585, valid at runtime on the version KiCad bundles; no new imports beyond `helpers`.
- **Settings sections are safe to extend** — every consumer of `settings["partselector"]` reads by explicit key (`partselector.py:273-282`, `:685`), and `SettingsDialog` binds its controls by name rather than iterating the section, so `column_widths` / `size` can't leak into the filter checkboxes.
- **Every close path funnels through `EVT_CLOSE`** — the X button, `quit_dialog`, and the Ctrl+W / Ctrl+Q / Shift+Esc accelerators all call `Close()`, and `_part_selector.Close()` in the main window's `quit_dialog` covers the parent-closes-first case. Nothing bypasses the save.

## 1. The part selector has no trailing spacer, so its last column is the stretched one

`datamodel.py:26-30` documents that the main window needed `TRAILING_SPACER_COL` specifically so the last visible column "doesn't stretch when the control is wider than the total column width." `PARTSELECTOR_COLUMNS` ends at `price` with no equivalent, so on platforms where wx reports the stretched width back from `GetWidth()`, `Price` gets persisted at window-width-minus-everything-else.

That's stable while the window stays the same size, but once it's saved the columns total the old window width — shrinking the selector afterwards gives a horizontal scrollbar instead of re-fitting. Since the PR notes native resizing wasn't exercised on-device, this is the case I'd check first. The same spacer trick would fix it.

## 2. Saved widths are keyed by a positional index, so adding or removing a selector column will misassign them

`get_column_widths` keys on `column.GetModelColumn()`, and in the selector that index is simply the position in `PARTSELECTOR_COLUMNS` (`for idx, column in enumerate(PARTSELECTOR_COLUMNS)`). Within a session that's sound — `COLUMN_INDEX`, `MODEL_COLUMN_TYPES` and `SORTABLE_COLUMN_INDEX_TO_DB` all derive from that same list at import, so they can't disagree with each other.

What changes here is durability: the index is now written to `settings.json` and read back by a *later build*. As far as I can tell this is the first positional index the plugin persists — `library.order_by` resets to `"LCSC Part"` on construction (`library.py:52`), and everything else in `settings.json` is bools and strings. So a release that inserts, removes or reorders an entry in `PARTSELECTOR_COLUMNS` silently shifts every user's saved widths by one: insert a column at index 4 and the old `params` width lands on `stock`, `stock`'s on `mfr`, and so on to the end of the list. That list has been edited several times recently (Std indicator, side column, params), so this isn't hypothetical — and the failure is invisible, since a wrong-but-plausible width just looks like the feature not working.

The main window is immune: its model columns are named constants in `datamodel.py` rather than positions.

Two ways to guard it:

- **Persist a stable key instead of the position.** `get_column_widths` is generic over wx columns so it can't reach `PartSelectorColumn.key`, but the call sites can supply one — pass an index-to-key mapping (`[c.key for c in PARTSELECTOR_COLUMNS]` in the selector, the inverse of `PartListDataModel.columns` in the main window) and key the saved dict on that. A couple of lines per call site, and it keeps a user's layout working across a column change instead of just not corrupting it.
- **Or version the saved blob and discard on mismatch** — store the column count or label list alongside the widths and drop the whole dict when it doesn't match what's on screen. Cheaper to write, but it throws away the user's layout every time the column list changes.

I'd take the first. Either way it's much cheaper now than later: once indices are on disk in a release, changing the scheme means carrying a migration for them.

One thing that is *not* an argument for keeping the index: I initially read the model-column keying as protection against users drag-reordering columns. There's no `DATAVIEW_COL_REORDERABLE` anywhere in the repo and wx's default column flags are `RESIZABLE` only — the `Std` column passing `flags=0` shows the flags are set deliberately — so the columns aren't user-reorderable and the index buys nothing over a stable key. (The reordering half of `test_column_widths_round_trip_json_by_model_column_across_dpi_and_reordering` is covering a state the UI can't currently reach. Harmless as defensive coverage, but worth knowing it isn't load-bearing.)

## 3. Unguarded `self.footprint_list` during teardown

`mainwindow.py:764` reads it before the `contextlib.suppress(AttributeError)` block that guards the logging handlers two lines below, and `get_part_details` guards `hasattr(self, "library")` for the same reason. A close after a partially-failed init now raises where it previously didn't.

## Nits

- Closing the main window with the selector open writes `settings.json` twice — the selector's `_on_close` calls `parent.save_settings()`, then `quit_dialog` saves again. Harmless, easy to collapse.
- The blank-title heuristic in `get_column_widths` works because the spacer's title is `" "`, but it couples "no header text" to "don't persist" — an icon-only header column later would silently stop persisting. A one-line comment pointing at `TRAILING_SPACER_COL` would make the intent readable.

z2amiller:
### Previous findings — all addressed

New — two things that probably should be addressed since they could lock someone out, and some nitpicks:

### Restored size isn't clamped to the display (new). 
partselector.py:575-581 checks only len == 2 and > 0. On Cocoa a saved [4000, 3000] opens clamped but is re-persisted as-is, because a clamped window reports IsMaximized() == True so _on_size never corrects it; [1, 1] opens a 1×1 window. Windows/GTK don't clamp at all. Clamp to wx.Display.GetFromWindow(self).GetClientArea() with a floor, and add IsFullScreen() to the _on_size guard.

This one seems pretty straightforward to fix and prevents someone getting stuck.


### Column widths have no upper bound (new).
window_layout.py:41; restoring 3000 into a 780 px control collapses the spacer and pushes columns off-screen. Clamp to the control's client width.

(This one is similar to the first one, where a bad value might hose the window and make it hard to recover)

More minor stuff, nonblocking:

### Every close now writes settings.json, non-atomically (pre-existing mechanism, new exposure).
save_settings truncates then dumps and load_settings is a bare json.load, so a write that fails mid-dump leaves a file the next launch can't parse, hidden by the new except OSError. Write to a temp file and os.replace, and wrap the save in try/finally so a non-OSError can't skip Destroy().

### Stale comment (new):
mainwindow.py:157-160 still says the file is only rewritten on a settings change.
Two plugin windows (pre-existing class, unconditional now): closing an unchanged window overwrites settings changed in another, since each holds its own snapshot. Noting, not asking.

Nits: PARTSELECTOR_COLUMN_KEYS is the inverse of COLUMN_INDEX and belongs in partselector_columns.py; the "don't persist" rule is spelled twice with string literals, so either a persist_width flag or a comment in window_layout.py that the last column must be an unpersisted spacer, since that invariant is what makes GetWidth() safe to save; to_dip mirrors HighResWxSize in helpers.py.

Unverified: on GTK IsMaximized() follows the asynchronous window-state event, so if the size event lands first _on_size records the maximized geometry and _on_maximize won't undo it. Worth one try on Linux if anyone can.

z2amiller (approved):
Looks like this got dirtied by #816 -- merge approved after deconflicting!

z2amiller (approved):
Rebase onto current main only; the conflict resolution in mainwindow.py is just the merged typing import and unquoted annotations under the existing from __future__ import annotations. Suite passes at 1372. Re-approving.

PR #819, opened 2026-09-06
fix(fabrication): export decimal CPL coordinates (#769)

### LINKED ISSUE #769: Component with small X value in footprint confuses JLCPCB importer
**Describe the bug**
Very small X / Y value in footprint gets exported to CSV as scientific notation, then the JLCPCB importer parses the coordinate value as its floating-point mantissa.

**To Reproduce**

place a component with X/Y coordinate of, say, 0.000005, then export. Exporter tool will create a field "5e-05" in the CPL-project.csv output file.

The JLCPCB importer will silently strip the 'e-05' from the CSV field and interpret the value as 5, placing the component at x=5mm.

# KiCad Version
Application: KiCad x86_64 on x86_64

Version: 10.0.3-10.0.3~ubuntu24.04.1, release build

Libraries:
	wxWidgets 3.2.4 GLX
	FreeType 2.13.2
	HarfBuzz 8.3.0
	FontConfig 2.15.0
	libcurl/8.5.0 OpenSSL/3.0.13 zlib/1.3 brotli/1.1.0 zstd/1.5.5 libidn2/2.3.7 libpsl/0.21.2 (+libidn2/2.3.7) libssh/0.10.6/openssl/zlib nghttp2/1.59.0 librtmp/2.3 OpenLDAP/2.6.10

Platform: KDE neon User Edition, 64 bit, Little endian, wxGTK, X11, wayland, KDE, KDE
OpenGL: AMD, AMD Radeon RX 7800 XT (radeonsi, navi32, LLVM 20.1.2, DRM 3.64, 6.17.0-29-generic), 4.6 (Compatibility Profile) Mesa 25.2.8-0ubuntu0.24.04.1, GLX 1.4

Build Info:
	Date: May 13 2026 23:54:14
	wxWidgets: 3.2.4 (wchar_t,wx containers) GTK+ 3.24
	Boost: 1.83.0
	OCC: 7.6.3
	Curl: 8.5.0
	ngspice: 42
	Compiler: GCC 13.3.0 with C++ ABI 1018
	KICAD_IPC_API=ON
	KICAD_USE_PCH=OFF

Locale: 
	Lang: en_US
	Enc: UTF-8
	Num: 1,234.5
	Encoded кΩ丈: D0BACEA9E4B888 (sys), D0BACEA9E4B888 (utf8)

chmorgan:
Small CPL coordinates currently reach csv.writer as floats, so 0.000005 mm is exported as 5e-06. Issue #769 reports that JLCPCB interprets the mantissa as the coordinate, moving the component by millimetres.

Format Mid X and Mid Y with six decimal places after position corrections and Y inversion. This prevents exponent notation while preserving KiCad's one-nanometre coordinate resolution.

Fixes https://github.com/Bouni/kicad-jlcpcb-tools/issues/769

Evidence:
- [JLCPCB CPL requirements](https://jlcpcb.com/help/article/pick-place-file-for-pcb-assembly) specify centroid coordinates in millimetres.
- [JLCPCB's EAGLE exporter](https://github.com/JLCPCB/jlcpcb-eagle/blob/master/ulps/jlcpcb_smta_exporter.ulp) writes decimal coordinates using %5.2f.
- [KiCad's coordinate units](https://docs.kicad.org/doxygen/base__units_8h.html) establish the precision retained by six decimal places in millimetres.

The JLCPCB examples support decimal formatting; they do not prescribe six decimal places or document exponent parsing. The exporter defect is reproduced locally; a live JLCPCB upload was not performed.

Validation:
- Tests-first commit: new regression module produced 13 expected failures and 4 passing controls before the fix.
- Full pytest suite after the fix: 467 passed; 5 file-manager tests could not start their localhost HTTP server inside the sandbox. Rerunning common/test_filemgr.py with localhost access passed all 10 tests, covering all 472 suite cases across the two runs.
- Ruff check and Ruff format --check passed for both changed files.
- git diff --check and commit hooks passed.

The history contains a tests-only commit followed by the two-line production fix.

### DIFF
--- a/fabrication.py
+++ b/fabrication.py
@@ -461,8 +461,12 @@ class Fabrication:
                         part["reference"],
                         part["value"],
                         part["footprint"],
-                        ToMM(position.x),
-                        ToMM(position.y) * -1,
+                        # Fixed-point millimetres follow JLCPCB's exporter:
+                        # https://github.com/JLCPCB/jlcpcb-eagle/blob/master/ulps/jlcpcb_smta_exporter.ulp
+                        # Six decimals preserve KiCad's 1 nm internal resolution:
+                        # https://docs.kicad.org/doxygen/base__units_8h.html
+                        f"{ToMM(position.x):.6f}",
+                        f"{ToMM(position.y) * -1:.6f}",
                         self.fix_rotation(fp),
                         "top" if fp.GetLayer() == 0 else "bottom",
                     ]
--- a/tests/test_fabrication_cpl_format.py
+++ b/tests/test_fabrication_cpl_format.py
@@ -1,4 +1,4 @@
-"""Keep CPL coordinates in decimal millimetres for GitHub issue #769."""
+"""Keep CPL coordinates in decimal millimetres."""
 
 import csv
 from pathlib import Path
--- a/tests/test_fabrication_cpl_format.py
+++ b/tests/test_fabrication_cpl_format.py
@@ -1,0 +1,143 @@
+"""Keep CPL coordinates in decimal millimetres for GitHub issue #769."""
+
+import csv
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+from .wx_harness import load, module, package_stubs
+
+
+class Point:
+    """Represent KiCad's integer nanometre coordinates."""
+
+    def __init__(self, x, y):
+        self.x = int(x)
+        self.y = int(y)
+
+    def __sub__(self, other):
+        """Subtract the board auxiliary origin."""
+        return Point(self.x - other.x, self.y - other.y)
+
+
+@pytest.fixture
+def generate_cpl(tmp_path):
+    """Run the real exporter with KiCad substitutes and temporary output."""
+    package = "_cpl_format_test"
+    stubs = package_stubs(package)
+    stubs["pcbnew"] = MagicMock(
+        ToMM=lambda value: value / 1_000_000,
+        FromMM=lambda value: round(value * 1_000_000),
+        wxPoint=Point,
+        VECTOR2I=Point,
+    )
+    helpers = f"{package}.footprint_helpers"
+    stubs[helpers] = module(helpers, get_is_dnp=lambda footprint: False)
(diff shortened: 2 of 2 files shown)


z2amiller:
Could this use the '.6g' modifier instead of the '.6f' modifier for the printf formatting?
That'd let numbers like "121.1" format as "121.1" instead of "121.100000".
(the 'g' truncates at repeated zeros after the decimal)

chmorgan:
@z2amiller `.6g` brings back scientific notation and rounds to six significant digits:

```python
for x in (121.1, 0.000005, 12.345678):
    print(f".6f: {x:.6f}  .6g: {x:.6g}")
```

```text
.6f: 121.100000  .6g: 121.1
.6f: 0.000005  .6g: 5e-06
.6f: 12.345678  .6g: 12.3457
```

I'd keep `.6f` to avoid the original bug and preserve coordinate precision.

z2amiller:
> @z2amiller `.6g` brings back scientific notation and rounds to six significant digits:
> 
> ```python
> for x in (121.1, 0.000005, 12.345678):
>     print(f".6f: {x:.6f}  .6g: {x:.6g}")
> ```
> 
> ```
> .6f: 121.100000  .6g: 121.1
> .6f: 0.000005  .6g: 5e-06
> .6f: 12.345678  .6g: 12.3457
> ```
> 
> I'd keep `.6f` to avoid the original bug and preserve coordinate precision.

Oh, that's annoying.  I guess if JLC accepts the trailing zeros in the CPL that's better than falling
back to scientific notation.

There's also decimal.Decimal which is in the stdlib, something like
```python
from decimal import Decimal

def clean_format(num):
    # Convert float to string first to avoid floating-point inaccuracies
    d = Decimal(str(num))
    # normalize() drops trailing zeros, and :f blocks scientific notation
    return f"{d.normalize():f}"
```

but idunno if that's gilding the lily.  If you think that's not worth the hassle,
we can merge this as-is.

z2amiller (approved):

chmorgan:
@z2amiller I bet they'd be fine with trailing zeros, I'll certainly know in the near future when I submit my next board.

PR #824, opened 2026-09-06
Seed settings.json from a shipped default_settings.json

### LINKED ISSUE #808: proposal: Create a default_settings.json file that gets copied to settings.json if empty.
When developing anything that touches the settings file, it gets dirtied with local changes.
We should be opinionated about good default settings that are good starting points for
new users, but also lets us .gitignore the settings.json files locally so they don't need to be
stashed/removed/whatever every time we're working on a PR.

Does this sound reasonable?  We'd have a default_settings.json, make sure it gets
packaged into the PCM, etc, and then copy it to the runtime settings.json if that file
doesn't exist.

z2amiller:
Fixes #808.

`settings.json` is tracked, so any development session that toggles a checkbox leaves a change to stash or revert, and switching branches trips over it. The file's history shows the result: the part-selector filters flipped in four unrelated PRs (`77780b3`, `5e56f5b`, `a15fc3d`, `ca8fcbb`), none of which were about settings.

This ships the defaults as `default_settings.json` and stops tracking the runtime `settings.json`, which `.gitignore` has listed since #742 but cannot ignore while the file is tracked.

### Merge rather than copy

On load, the stored settings are layered over the defaults section by section, so a setting added in a newer version arrives with its default instead of being missing. The result is written back only when it differs from what is on disk, so a settled file is not rewritten on every launch.

That replaces the default patching `load_settings()` had grown: `subtract_mask_from_silk`, `simplify_stock`, `stock_concern` and the two part-preference switches are entries in `default_settings.json` with the same values. The `highlight_matches` rename runs against the stored file *before* the merge, so a stored `false` is not masked by the default `true`, and the `force_drc` / `fill_zones` repair runs after it.

A settings file that cannot be parsed is moved to `settings.json.damaged` rather than overwritten, so the plugin still opens and the damaged file survives for recovery; if it cannot be moved aside, that error surfaces instead, so the write-back never replaces the only copy. A file that cannot be read at all, because of permissions or another program holding it, is left alone and the error surfaces as it always has, since its contents may be fine.

`core/settings_defaults.py` imports neither wx nor pcbnew, so the whole resolution path is unit-tested outside KiCad.

### Defaults

Two shipped defaults change:

- **The part selector includes Basic, Preferred and Extended.** `ComponentTranslator.library_type()` returns those as mutually exclusive labels, so the shipped `preferred`-only default hid *every Basic part* from a fresh install's search results. The in-code fallbacks in `partselector.py` already default all three to `True`.
- **The BOM-estimator settings are shipped** like the rest. They previously existed only as call-site fallbacks.

Existing installs keep what they stored, by design. For installs updated through PCM that is usually the preferred-only filter, because the 2026.04 releases shipped it in `settings.json` and each update wrote it over the user's file. This change does not rewrite anyone's settings to correct that. Its point is that a new install starts from defaults chosen on purpose, rather than from whatever `settings.json` held when someone committed it after testing or simply using the plugin.

Three call sites fall back to a different value than the one shipped: `force_drc` (`False`), `lcsc_priority` (`True`) and `order_number` (unset). A settings file old enough to lack one of those keys now resolves to the shipped value, which is what releases have carried since each key was added.

### Tests

`tests/test_settings_defaults.py` scans the plugin's sources for setting reads and fails on any that `default_settings.json` does not cover. It tolerates newlines inside the call, because `ruff format` wraps it once a setting name is long enough, which is exactly the case the guard exists for. Settings read through a section the code has already pulled out of the dict are listed in the test by hand.

The guard found a live bug, fixed here: `PartSelectorDialog.get_highlight_text()` still read the pre-rename `partselector.highlight_matches`, which loading settings moves out of every stored file, so unchecking "Highlight search matches" stopped highlighting in the main window but not in the part selector's own results.

The settings persistence and stock-concern tests lay out `default_settings.json` beside the `settings.json` they load, as an installed plugin has it, since that is where the defaults they check come from. Their scenarios and expected values are otherwise unchanged:

- the orphaned-temporary-file checks allow for the defaults file;
- the `saved_settings` document starts from the shipped defaults, so loading it adds nothing;
- the simplify-stock expectation includes the rest of the shipped `general` section.

### Packaging

The archive carries `default_settings.json` in place of `settings.json`, which is what lets #822 work. On an update, PCM's `deletePackageDirectories()` keeps the files matched by `keep_on_update`, and `extract()` then writes every archive entry into the plugin directory without consulting that list (`kicad/pcm/pcm_task_manager.cpp`). While the archive shipped a `settings.json`, that extraction overwrote the one just kept. `plugins/[^/]+/settings\.json$` does not match `default_settings.json`, so each update refreshes the defaults.

### Git checkouts

This deletes the tracked `settings.json`, which affects development checkouts and anyone running the plugin from a git clone.

- Normal use of the plugin rewrites `settings.json`, so a clone usually has local changes to it and `git pull` refuses to update; stash, pull and pop ends in a modify/delete conflict instead. Move the file aside, pull, and move it back: it is then the git-ignored runtime file.
- A branch that still tracks `settings.json` keeps the old friction until it is rebased past this, and checking one out replaces the runtime `settings.json` with the tracked copy, because git treats ignored files as expendable.

### Open PRs that add settings

#784 adds `gerber.export_schematic_pdf`, `gerber.export_pcb_layer_pdfs` and `general.auto_search_keywords`, and edits `settings.json`. Whichever of the two lands second adds those three to `default_settings.json`; the guard test fails until they are there.

### Verification

- 1904 tests pass (1891 on `main` plus 13 new).
- `ruff format --check` is clean on every touched file; `ruff check`, markdownlint and deadcode report nothing that `main` does not.
- The upgrade path was run against the tracked `settings.json` from `main` and from before #809: every stored value survives, the missing keys are seeded, and a second load writes nothing. A stored `highlight_matches: false` migrates as `false`, and `force_drc` without `fill_zones` is repaired.
- A `settings.json` the process is not allowed to read raises, and so does a damaged one that cannot be moved aside; either way the file stays in place with its contents intact.


chmorgan:
looking forward to this PR :-) hitting the settings conflict right now just changing branches

chmorgan (approved):
Reviewed 31ba9e22aa024257cfe0f9199f3c95f1008e9f79. No actionable issues found in settings migration and recovery, packaging and upgrades, or UI consumers.

Validation: 212 focused tests passed; scoped Ruff lint and formatting checks passed. Additional probes passed for first-run persistence, reopening without rewrites, legacy highlight migration, newly added defaults, preservation of saved choices, and the changed highlight getter.

Native KiCad repaint behavior and a PCM update were not exercised; packaging was reviewed through source inspection.

Issue #1, opened 2021-08-09
Merge feature cpl_rotations_db.csv of JLCKicadTools 


bvernoux:
It will be great to merge the feature cpl_rotations_db.csv of https://github.com/matthewlai/JLCKicadTools as I have do not found any way to fix the rotation in cpl in the actual plugin

Bouni:
I'll see if I can include that as well!
Thanks for the input 😄

Bouni:
@bvernoux I've integrated the fix as you proposed, would you mind testing if the result is valid?

bvernoux:
Thanks for the fast update
I will compare with https://github.com/matthewlai/JLCKicadTools using my own cpl_rotations_db.csv (I have added few parts not existing)

Bouni:
I did a really quick test with a dummy PCB and the outcome seems perfectly fine!

![DeepinScreenshot_select-area_20210810211932](https://user-images.githubusercontent.com/948965/128921254-f93214c2-4e29-43f7-a988-ad8a7f6e6171.png)

Bouni:
> Thanks for the fast update
> In fact I have checked the code modification [6de05c9](https://github.com/Bouni/kicad-jlcpcb-tools/commit/6de05c9b75860118101940d37170f480a7517fe6) but what I'm doing is I modify this file feature cpl_rotations_db.csv with my own modifications (as there is always new part not in this file) so it will be better to search that file in current directory for example and if it does not exist to download it

I guess you're right. Maybe we should look for a local file, if not present, download it from GitHub!?

What do you think about that?

bvernoux:
Yes it will be perfect like that I could check with my different design with expected results

Bouni:
@bvernoux Ok, I've implemented the use of the local file. Highly appreciate that you test it!
Even greater would be if you pretend to let one or more of your boards be made by JLC and check what the preview shows.
That way we could check if everything is correct now (screenshots would be awesome 😎 )

bvernoux:
I do not know why but with my KiCad 5.99 Nightly (r22289) on Windows 7 kicad-jlcpcb-tools plugin does not start and it is not displayed
Does your plugin is compatible with Python 2.7 or requires Python 3.x ? (as on KiCad it is a bit of a mess and only latest KiCad 5.99 support Python3 IIRC but it does not support anymore Windows7 ...)

bvernoux:
[...]

PR #826, opened 2026-09-09
Add compact stock display and grouped stock concern highlighting

chmorgan:
Add two independent, default-enabled Settings options:

- **Simplify stock** displays `22095 → 22 k`, `8880000 → 8.8 M`, and `7260 → 7.2 k` in the main list and part selector. Formatting truncates without changing exact sorting, assignment, or storage values.
- **Highlight stock concern** uses bold amber Stock text when known supply is strictly below `10 × populated references sharing an LCSC × estimator board quantity`, or availability is unknown. Missing stock displays `?`; unassigned rows remain blank. BOM-excluded and DNP parts do not count, while hand-placed BOM parts excluded from CPL and hidden rows still contribute.

The concern uses the saved estimator quantity even when its panel is hidden. Its existing minimum/default of five boards is unchanged. Both settings apply immediately and survive reopening. Normal cells retain native backgrounds and selected cells retain native selection colors.

## Implementation

- Restore blank stock before zero in ascending order, with descending reversed. Numeric values continue sorting by exact stock even when their compact labels match.
- Share stock presentation/comparison/notification behavior between both models.
- Share raw catalog details across population, concerns, estimation, and assignments. Warmed BOM/filter/enrichment/quantity events do not repeat catalog queries; confirmed missing records are cached distinctly from transient failures.
- Publish selected stock only after assignment succeeds, updating all matching visible rows together. Malformed raw stock remains unknown in persistent assignments.
- Refresh stock, concerns, prices, and open/new selectors consistently after catalog changes. Serialize source switches during downloads; source and attempt identities reject obsolete completions. Publication and failure recovery clear stale snapshots and preserve initial saved preferences/enrichment.
- Keep capture scripts and images outside the proposed source tree.

## Catalog and event handling

- Suppress synchronous wx events while rebuilding selector filters, then search once after the final values are restored. Repeated unavailable resets and recovery retain working handlers; selecting the display-only **All** category does not query a nonexistent category key.
- Settings edits apply catalog configuration without starting a catalog download, including deferred source changes and recovery after failed initialization. The existing Update action and opening flow retain their download behavior. Path configuration continues to apply immediately.
- Validate catalog consumers and database integrity without creating or migrating storage. A failed transfer preserves an intact usable catalog; missing, corrupt, or incomplete replacements remain unavailable. Optional missing update metadata remains supported. Integrity scans run in the download worker; UI publication uses a smaller schema/query check.
- Restore the visible storage-error banner and disabled actions after directory or publication failures. Invalid paths and malformed update timestamps cannot remain initialized; valid Settings retries restore the controls and current catalog.
- Remove the unreachable `refresh_stock()` method. Cache invalidation tests now exercise the real Update handler, Library worker, completion events, and source switching, including formerly missing records.

## Assignment feedback and generation recovery

- Paste LCSC, Apply part preferences, and late part-selector assignments now log one actionable warning when project storage or the selected catalog is unavailable. Refused actions leave assignments, board fields, displayed rows, and catalog downloads unchanged; background refreshes stay quiet.
- Recovery through Update or Settings initializes the real Fabrication object before clearing the storage error and enabling Generate. Output-directory failures retain the banner, diagnostic tooltip, and disabled Generate until a successful retry. Existing generator instances are reused.
- Save preferences and Clear assignment remain available without a catalog. Assignment retries work after download or switching back and survive reopening. Selecting an unfetched replacement catalog can restore project-only generation without starting a download.

## Screenshots

These are native macOS controls with deterministic sample inventory, using the production models and Settings controls. They are not a full KiCad board session or live supplier inventory. The two checkboxes normally live in **Settings**; they are shown above the fixture table only for comparison. The fixture uses five boards, so three matching references warn at 149 stock and clear at 150. Unknown stock is shown explicitly as `?`.

### Simplify stock only

Simplify stock enabled; concern highlighting disabled

### Highlight stock concern only

Concern highlighting enabled; Simplify stock disabled

### Both options enabled

Both options enabled with native row backgrounds retained

## Acceptance evidence

| Requirement | Evidence |
| --- | --- |
| Compact examples and raw sorting/storage | `test_stock_display`, `test_stock_model_revision`, and real selector/SQLite assignment workflows cover exact examples, magnitude boundaries, both modes/directions, equal compact labels, blanks and zero, and unchanged raw values. |
| Defaults, immediate toggles, and reopening | Real constructors/event handlers with stateful wx doubles and atomic-settings tests cover missing-key migration, explicit false, open selectors, durable reopening, and failed writes. |
| Grouped demand and board quantity | Policy/controller cases cover strict 4,999/5,000 boundaries for 5 copies × 100 boards, live BOM/DNP eligibility, hidden POS-excluded parts, and quantity spin/text handlers with a hidden estimator and persisted quantity. |
| Unknown supply remains a risk | Missing/invalid stock concerns eligible assignments, has a visible label, and persists as unknown rather than zero across selector/preferences paths. |
| Cached inventory remains coherent | Real estimator/controller/model workflows count catalog calls, exercise confirmed misses/transient recovery and defensive copies, and verify assignment failure plus sibling updates after success. |
| Catalog switching/download recovery | Real Library/SQLite workflows cover repeated path edits, deferred source changes after both worker outcomes, opening and explicit Update, failed transfers with intact catalogs, corrupt replacements, attempt identity, malformed metadata, invalid saved paths, visible errors and valid retry. |
| Selector event ordering | Constructor-bound stateful controls reproduce synchronous Clear events; native macOS probes cover repeated invalidation, one final search, recovery, All selection, stopped timers and handler restoration after exceptions. |
| Unavailable assignment feedback and retry | `test_stock_assignment_feedback` exercises Paste, Apply preferences, and selector events through missing catalog/path changes, pending source switches, and absent Library state. Actual Library/Store/SQLite checks verify one actionable warning, no mutation/download, successful retry and durable reopening, quiet automatic work, single SQL-error warnings, and preserved Save/Clear behavior. |
| Generation recovery and failed retries | `test_stock_fabrication_recovery` runs production initialization/event handlers with real Library, Store, SQLite, and Fabrication constructors. Corrupt/malformed catalogs recover through Update or Settings, including selection of an unfetched catalog. Both output-directory failure orders and population-read failure retain disabled generation until repair. Actual placement, consistency, and zone preflight runs through Generate and deliberately stops at DRC. |
| Native appearance | Native fixture checks cover all four flags, unknown labels, clearing, selected rows, and scrolling. Foreground-only attributes preserve normal row styling; light/dark palette regressions remain covered. |

## Validation

- History cleanup retains **9 feature and test commits**, including the spacing adjustment. Follow-up catalog repairs are folded into their introducing implementation and regression-test commits; the separate feature/test commits remain. Review-only images, capture tooling, and local guidance are absent throughout the branch history. All product and test changes are preserved. The branch is rebased onto `d5f506d`, with `origin/main` synchronized to `upstream/main`.
- New follow-up coverage: **39 cases**. Baseline replay gives **26 expected failures and 13 passing compatibility cases**, with no fixture/setup errors, before the fixes; compatibility cases cover quiet background work, SQL failures, project-only actions, and failed population recovery.
- Final complete suite after rebasing onto current main: **1,890 passed** using the project environment.
- Ruff lint/format checks, Python 3.9 syntax parsing for all 23 Python files in the PR, and diff whitespace checks passed.
- Acceptance tracing, independent requirements review, and final patch/commit review completed. The assignment and fabrication test workstreams were independent; review added Settings-to-unfetched-catalog and population-failure coverage. Legacy assignment fixtures explicitly represent prior Fabrication initialization; the new recovery fixture opts out and requires the real constructor.
- Previously recorded native macOS checks cover selector Clear-event ordering/recovery, both models' exact row sorting in both directions with simplification on/off, and directory-picker change events for nonexistent typed path prefixes. The three comparison captures were regenerated on the current head to show spaced compact labels; native option-event, row-attribute, scrolling, and Settings-reconstruction checks passed again.

The new recovery tests bypass the main dialog constructor while executing real initialization methods, storage, Fabrication construction, and generation preflight. wx/KiCad boundaries and download transport are doubled; DRC and exports deliberately do not run. Worker event ordering is covered with held workers and captured completion events; real concurrent-thread timing was not exercised. Native evidence is limited to macOS fixture controls; full KiCad board integration and native Windows/Linux/light-mode rendering remain outside this session's evidence. No build was required or run. Exact supplier details remain available in Part Details.


z2amiller:
A couple design questions/concerns first, then some nitpicks:

UI Design changes:

1. Stock sorting changes for unmapped or zero-stock items -- I think these end up being opposite of the current behavior.  Just wanted to check if that was intentional and whether or not it makes UI sense.
2. Was ignoring the board count on purpose? A 100-board order with a 5-per-board part should probably warn on less than 500*fudge_factor stock.  (Maybe not a huge deal since Idunno how many people plug the number of boards into the bom estimator)

Things that feel like blockers:

1. Parts absent from the catalog don't get flagged.  Since the default catalog discards long-out-of-stock items, if it's missing from the parts database it's more likely to have zero stock.  Not sure if we should assume zero on those, but probably we should flag them as stock risks if they're not in the catalog.  (Also things that fail to parse, etc, should flag as risks)
3. Full pass per BOM event - recompute_stock_concerns can run on the main UI thread and foreaches the parts thru a SQL lookup.   We should probably cache these - so it sucks a little on board open but doesn't hammer a bunch of SQL lookups every time something changes.  I think both the estimator and `partlist_data_model.data` already has a cache.  I suppose if there are times where you KNOW the cache is wrong (maybe the enrichment step gets fresh stock numbers? I don't remember. Or a new library download) there could be either a separate method or a `rebuild_cache=True` flag or something to take the slow path.

Minor:
1. A library switch can leave things in an inconsistent state - it probably wants to repop footprints (and throw out any caches)  -- calling `populate_footprint_list` would fix at least some of the inconsistency, maybe something else needed to blow away caches though.
2. docs/stock-feature-captures.md tells a contributor to run myenv/bin/python (no such environment in the repo) and mentions "Codex's native app capture surface"; the 405-line macOS-only harness and ~750 KB of PNGs have no precedent in the tree (earlier PRs attached screenshots to the PR). Suggest keeping the images in the PR description and dropping the doc, or rewriting it for a contributor. The three comparison images also show the two checkboxes reparented from the settings dialog into the fixture window above the table; on first read they look like new main-window controls, so a caption saying they live in Settings would help.
4. set_simplify_stock, the GetValue stock branch and the Compare stock branch are triplicated across the two models, differing only in the column literal; the direction fix for #1 has to land twice.

chmorgan:
> A couple design questions/concerns first, then some nitpicks:
> 
> UI Design changes:
> 
> 1. Stock sorting changes for unmapped or zero-stock items -- I think these end up being opposite of the current behavior.  Just wanted to check if that was intentional and whether or not it makes UI sense.
> 2. Was ignoring the board count on purpose? A 100-board order with a 5-per-board part should probably warn on less than 500*fudge_factor stock.  (Maybe not a huge deal since Idunno how many people plug the number of boards into the bom estimator)
> 
> Things that feel like blockers:
> 
> 1. Parts absent from the catalog don't get flagged.  Since the default catalog discards long-out-of-stock items, if it's missing from the parts database it's more likely to have zero stock.  Not sure if we should assume zero on those, but probably we should flag them as stock risks if they're not in the catalog.  (Also things that fail to parse, etc, should flag as risks)
> 2. Full pass per BOM event - recompute_stock_concerns can run on the main UI thread and foreaches the parts thru a SQL lookup.   We should probably cache these - so it sucks a little on board open but doesn't hammer a bunch of SQL lookups every time something changes.  I think both the estimator and `partlist_data_model.data` already has a cache.  I suppose if there are times where you KNOW the cache is wrong (maybe the enrichment step gets fresh stock numbers? I don't remember. Or a new library download) there could be either a separate method or a `rebuild_cache=True` flag or something to take the slow path.
> 
> Minor:
> 
> 1. A library switch can leave things in an inconsistent state - it probably wants to repop footprints (and throw out any caches)  -- calling `populate_footprint_list` would fix at least some of the inconsistency, maybe something else needed to blow away caches though.
> 2. docs/stock-feature-captures.md tells a contributor to run myenv/bin/python (no such environment in the repo) and mentions "Codex's native app capture surface"; the 405-line macOS-only harness and ~750 KB of PNGs have no precedent in the tree (earlier PRs attached screenshots to the PR). Suggest keeping the images in the PR description and dropping the doc, or rewriting it for a contributor. The three comparison images also show the two checkboxes reparented from the settings dialog into the fixture window above the table; on first read they look like new main-window controls, so a caption saying they live in Settings would help.
> 3. set_simplify_stock, the GetValue stock branch and the Compare stock branch are triplicated across the two models, differing only in the column literal; the direction fix for [Merge feature cpl_rotations_db.csv of JLCKicadTools  #1](https://github.com/Bouni/kicad-jlcpcb-tools/issues/1) has to land twice.

This is good feedback. Wish I had caught this stuff sooner, esp the AI adding images to the PR, that's nuts and I added an agents comment to my local project agents so it won't do that nonsense anymore.

z2amiller:
Getting close - two blockers, both event related:

1. Events are annoying, partselector.py:752 calls self.category.Clear(). On real wx that fires EVT_TEXT synchronously, even on an empty CB_READONLY box, and EVT_TEXT is bound to update_subcategories (:445), whose new first lines are if not self.parent.is_catalog_available(): self.refresh_catalog(); return (:807). So refresh_catalog → Clear → update_subcategories → refresh_catalog → (etc)

2. Events are annoying part 2: A Settings change now starts a full download, per keystroke. _apply_library_settings (:1931) ends the UPDATE_NEEDED branch with self.library.update(); main only called refresh_library_config() there. The data-path control is a DirPickerCtrl with DIRP_USE_TEXTCTRL bound to EVT_DIRPICKER_CHANGED, which on wxOSX fires on every text change including non-existent prefixes. Typing a path creates corrections.db/mappings.db in the first creatable prefix and starts a multi-hundred-MB download into it with no prompt; when that finishes, _catalog_switch_pending starts a second one for the real path. Leave downloads to the Update button and the open-time flow.

Smaller annoyances around error handling - not blocking:
1. _finish_download (library.py:1235) sets UPDATE_NEEDED on any failure without checking whether partsdb_file is still valid, and download_finished (mainwindow.py:1092) then calls _clear_catalog_views() -- this breaks the prices until a successful download or plugin restart, even if the remaining catalog is still valid.  (a fairly edge case TBH)
2. main routed library-switch errors through _set_project_storage_error (banner + tooltip + disabled controls); head logs a warning and clears views (:1932). And _apply_library_config assigns datadir/partsdb_file before setup() can raise, so library.state stays INITIALIZED against a directory that does not exist.

Full on nipick: 
1. refresh_stock is unreachable:  mainwindow.py:879 has no Bind, tool or menu; only three tests call it, one asserting that "an explicit refresh should discover a previously missing record", which no user can do. Confirmed misses are cached for the life of the snapshot, so a part absent at open time stays ? until restart. Wire it (with a library is None guard) or drop it and the tests.

chmorgan:
@z2amiller Sorry for the messy review rounds and the extra work this has put on you. I should have caught the wx event behavior and the incorrect test expectations before asking you to review again. Thanks for spelling out the failures so clearly.

The fixes are now in three separate commits, ending at 5f5ab606:

- Selector filter resets suppress synchronous events and run one final search. Repeated unavailable resets, recovery, and selecting **All** are covered.
- Settings changes no longer start catalog downloads, including deferred source changes and recovery after failed initialization. Downloads remain tied to the Update action and opening flow.
- Failed downloads retain a usable existing catalog. Read-only validation rejects corrupt or incomplete replacements, including missing columns used by selector searches.
- Invalid paths and malformed metadata restore the visible error banner and disabled actions; selecting a valid path restores operation.
- Removed the unreachable `refresh_stock()` method. Its coverage now goes through the actual Update handler, Library worker, and completion events.

The reported failures were reproduced before their fixes. The full suite passes **1,850 tests**, with Ruff lint/format and Python 3.9 syntax checks passing. Native macOS control probes verified selector event handling and the directory picker's per-keystroke notifications. Independent requirements and final-commit reviews are complete; the PR description includes the acceptance evidence and the approved screenshots.

Full KiCad integration, Windows/Linux native behavior, and real concurrent-worker timing remain unverified. The worker tests use held workers and captured events.

Could you take another look when you have a chance? Sorry again for the churn, and thank you for your patience.

z2amiller (approved):
A couple small fixes, more or less optional - I'll approve so you can merge,
or happy to do another round of review.

1.  Assignments silently do nothing with an undownloaded library.  I think that this is existing but becomes more visible because a library switch no longer auto-downloads: pick any library you have not fetched and Paste LCSC or Apply part preferences write nothing, log nothing, and the toolbar stays enabled. Main logged a failure on the same action. See mainwindow.py:1176.
2. Generate breaks after a catalog recovery. Fabrication setup runs only after library setup in init_data, and library setup now raises for any catalog that fails the new probe. The recovery then succeeds too well: Update completes, the banner clears, Generate is enabled, and pressing it fails with an attribute error until the dialog is reopened. Reproduced end to end by two independent agents. Since the fabricator is mocked, it doesn't catch this particular case.   Reachable from an interrupted extraction, since the unzip writes in place. See mainwindow.py:752.  Again -- probably quite a dusty corner case but worth fixing if you have the time (or tokens :-) )

chmorgan:
Thanks for catching both follow-ups, @z2amiller, and sorry for the extra review cycles. Fixed in 8155ec3, with separate regression coverage in 67fa280:

- Paste LCSC, Apply part preferences, and late selector assignments now give one actionable warning when storage or the selected catalog is unavailable. They preserve assignments and do not trigger downloads. Retrying after recovery works and survives reopening; Save preferences and Clear remain usable without a catalog.
- Update/Settings recovery now constructs Fabrication before clearing the error and enabling Generate. Output-directory failures keep the error visible and generation disabled until repair. Tests run the real Fabrication constructor and actual placement, consistency, and zone preflight, including corrupt/malformed catalogs and repeated failed recovery.

Validation: **1,889 tests passed**; the 39 new cases replayed against the previous head produce 26 expected failures and 13 passing compatibility cases. Ruff lint/format, Python 3.9 syntax, acceptance tracing, independent requirements review, and final commit review passed. The PR description now includes this evidence and retains the three approved screenshots.

The recovery tests double wx/KiCad boundaries and stop before DRC/export; full native KiCad generation remains unverified in this session.

Could you take another look? I am re-requesting your review on the updated head.

z2amiller (approved):
Approved - can you do one more thing, and squash before merge?
Specifically, there's a commit that adds the PNGs and then one that
removes them, so squashing before merge will keep those blobs out
of history.

chmorgan:
Yes of course. Let me remove those images. I thought they were deleted from
history with the last push.

chmorgan:
@z2amiller Sorry again for the extra churn, especially claiming the image cleanup was complete when the earlier commits still contained the files.

The branch history is now rewritten to 9 commits:

- Screenshots, capture tooling/notes, and the local agent instructions are absent from every PR commit, and their recorded blobs are unreachable from the rewritten branch.
- Follow-up catalog fixes are folded into the implementation commit that introduced those paths; their regression tests are folded into its test commit. The separate feature/test commits remain.
- The branch is rebased onto current upstream main (`d5f506d`), and the fork's main is synchronized.
- Compact labels now include a space (`22 k`, `8.8 M`, `7.2 k`), with the requested BIPM citation in the code and commit. The PR screenshots have been refreshed to match.

Independent feedback and history review found no unresolved reported issues. The final source tree is identical to the pre-consolidation tree that passed 1,890 tests; the history rewrite changed no code. Native macOS option, clearing, scrolling, reopening, and exact-sorting checks passed on the published source. The PR description retains the acceptance evidence and the full KiCad, other-platform, export, and worker-timing limits.

Could you review the updated head again? Thank you for your patience.

z2amiller:
No worries on the review passes, it's a big (and useful) change and it's important to get it right!

PR #827, opened 2026-09-10
Fix Generate doing nothing on KiCad builds made with SWIG 4.5 (SwigPyIterator.next() removed)

lucaci32u4:
# Bug descriprion

On KiCad builds generated with SWIG 4.5.0 (Arch `kicad 10.0.5-3` from 2026-08-28 onwards), clicking **Generate** does nothing.

SWIG 4.5.0 dropped `SwigPyIterator.next()` together with the rest of its Python 2 support (swig/swig#3201). KiCad's own SWIG interface adds a hand-written `__iter__` to the `DRAWINGS`, `TRACKS` and `VECTOR_SHAPEPTR` wrappers so it can `Cast()` every `BOARD_ITEM` to its concrete type, and that code still calls `it.next()`. 

On these builds iterating `board.Drawings()` and `board.GetDrawings()` raise:

```
  File "/usr/lib/python3.14/site-packages/pcbnew.py", line 22180, in GetDrawings
    def GetDrawings(self):            return list(self.Drawings())
  File "/usr/lib/python3.14/site-packages/pcbnew.py", line 13555, in __iter__
    item = it.next()  # throws StopIteration when iterator reached the end.
AttributeError: 'SwigPyIterator' object has no attribute 'next'
```

`Footprints()`, `Zones()` and `Pads()` are not affected, they go through SWIG's stock iterator.

The plugin hits this in `count_order_number_placeholders()`, which runs on every Generate. The call sits outside any `try`, so the exception escapes the wx event handler and is not shown in the built-in log console.

# Changes

- `footprint_helpers.iter_board_items(container)`: iterate normally, and if that raises the `AttributeError`, fall back to `len()` which still works.
- `mainwindow.count_order_number_placeholders()` iterates `board.Drawings()` through that helper instead of calling `GetDrawings()`. `Drawings()` is the C++ method the `GetDrawings()` wrapper itself calls and exists on every KiCad version the plugin supports.
- The hook-only placeholder count in `generate_fabrication_data()` now runs inside `run_generation_step("Counting order/serial placeholders", ...)`, so a failure there is reported against the right step.
- `standalone_impl.BoardStub` exposes `Drawings()` instead of `GetDrawings()`, and the test harness stub for `footprint_helpers` gained the new symbol so the existing mainwindow tests keep loading.

# Testing
- Test suite suite: 1579 passed, 17 warnings.
- Tested on 6 of our internal KiCad projects, generation is now fixed on machines with newer KiCad versions.

Alexandru Lucaci
Alacrity Education

### DIFF
--- a/footprint_helpers.py
+++ b/footprint_helpers.py
@@ -1,5 +1,6 @@
 """Helpers for reading and mutating KiCad footprint and board state."""
 
+from collections.abc import Iterator
 import re
 from typing import Any, Optional
 
@@ -72,6 +73,19 @@ def get_valid_footprints(board):
         if re.match(r"[\w\d-]+", fp.GetReference()):
             footprints.append(fp)
     return footprints
+
+
+def iter_board_items(container: Any) -> Iterator[Any]:
+    """Yield the concrete board items held by a pcbnew container."""
+    try:
+        items = list(container)
+    except AttributeError:
+        items = []
+        for index in range(len(container)):
+            item = container[index]
+            cast = getattr(item, "Cast", None)
+            items.append(cast() if callable(cast) else item)
+    return iter(items)
 
 
 def get_bit(value, bit):
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -65,6 +65,7 @@ from .footprint_helpers import (
     get_exclude_from_pos,
     get_is_dnp,
     find_lcsc_assignment_text,
+    iter_board_items,
     set_lcsc_value,
     toggle_exclude_from_bom,
     toggle_exclude_from_pos,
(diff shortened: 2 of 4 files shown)


dkm:
FWIW, I've applied this change over today's main branch, and it fixed the issue (using Debian's 10.0.6+dfsg-1 package). Thanks!

z2amiller (approved):
Nothing really blocking - but could you expand the commit message with the body of this PR?
(The PR description is great - but that gets lost in the squashed commit, leaving the PR as the only record
of why the change was made)

Also it sounds like they may have fixed this upstream (in KiCad itself, in 10.0.7?) but this is an easy fix
and removes this breakage path.

lucaci32u4:
I have amended the commit to contain the description content inside the commit message.

z2amiller:
Thanks for the fix!

PR #830, opened 2026-09-12
Explain library type fees with a wxWidgets hover table

chmorgan:
Hovering over the main window's Type cells displays a two-column table explaining library types and feeder loading fees. The table covers Basic, Preferred, Extended, and blank Type cells; the existing Standard-only explanation remains available elsewhere on the row.

The popup uses wxWidgets controls, sizers, system tooltip colors, and the window font. It adds no columns or dependencies and makes no database or saved-layout changes. The branch is rebased onto `main` at `19afa41` and preserves the merged frame/panel, selection/filter, settings, and stock/catalog behavior. Implementation and test commits are separate, with each test commit following its implementation.

The wording follows JLCPCB's [assembly pricing](https://jlcpcb.com/help/article/pcb-assembly-price) and [assembly FAQ](https://jlcpcb.com/help/article/pcb-assembly-faqs). No currency amounts are embedded.

### Review fixes

- Keep both help types working when a modeless dialog remains behind the active main window. On macOS and GTK, wx's [global window lookup](https://github.com/wxWidgets/wxWidgets/blob/v3.3.3/src/common/utilscmn.cpp#L1294-L1354) searches window creation order without checking native stacking order. Hover eligibility now uses the active owner, the table body's client bounds, native cell hit-testing, and a separate screen-rectangle exclusion for the fee popup.
- Exercise timer scheduling and coordinate conversion through stateful doubles. Advancing time dispatches only due events from running timers, respecting interval, repetition, source, and shutdown. Hit-testing depends on distinct screen, control, and body coordinates.
- Connect the real controller and fee-table factory in one workflow, verifying the displayed content and list ownership through dismissal and a fresh controller.
- Retain the earlier fixes that destroy partially allocated popup windows after construction failures and clear help before public table refreshes, including unavailable catalog/project data.

### Acceptance evidence

| Requirement | Evidence |
| --- | --- |
| Help survives background dialog overlap | Four new regressions failed against the original implementation for missing Type/Standard help, and pass after the fix. A native macOS fixture reproduced the original failure with the exact old controller and passed the fix with both window creation orders. |
| Foreground dialogs suppress help; returning restores it | Stateful tests cover both help types. Native activation transitions clear visible Type help and restore it at unchanged pointer coordinates. |
| Closing and reopening does not retain stale hover state | Closing, destroying, and recreating the native fixture's auxiliary dialog preserves both help paths with the same controller. Native main/table destruction stops the old timer; a new controller shows Type help. |
| Stationary hover and later cleanup use real scheduling | Tests deliver scheduled timer events without mouse motion. Native execution observes a repeating 100 ms timer, delayed display, and timer shutdown. |
| Correct coordinate space and table boundaries | Nonzero control/body origins and horizontal scrolling affect test hit results. Native macOS header, scrollbar gutter, and outside-table checks show no help. Popup positioning receives the screen-space pointer. |
| The shown popup is the fee table owned by the list | Real controller/factory integration checks all ten text cells, list parent, font/scale, fitted size, display, child destruction, and a fresh controller. Native execution confirms the popup's list parent and position. |
| The popup excludes underlying cells | Tests cover Type and Standard cells under its rectangle. A native probe deliberately moved the real popup over a stationary pointer and observed timer-driven dismissal. This was artificial relocation, not natural screen-edge flipping. |
| Existing content, refresh, close, and failure behavior | Exact-text, constructor, metadata, model-reset, empty-table, and failure-cleanup tests pass. Explicit late timer events cannot recreate help after stop or owner destruction. |
| Rebased Frame → panel → table hierarchy | A native macOS fixture against `30b4550` passed with both auxiliary/main window creation orders. It used the production tooltip modules and unchanged extracted Layout, Standard-help, and close handlers. Real controls verified active-frame lookup, timer-delayed Type help, all ten labels, table ownership and screen position, native Standard tooltip installation/clearing, activation suppression/restoration, table destruction, and both help paths after reopening. Pointer position/buttons were injected; physical input delivery was not verified. |
| Compatibility and scope | Both changed files parse with Python 3.9 grammar. No storage schema, dependencies, columns, or platform-specific calls were added. |

- Rebased suite: **1,983 passed** in the initial run; five download fixtures could not bind local HTTP sockets. Rerunning `common/test_filemgr.py` with socket access passed all **10 tests**, including those five. All **1,988 collected tests** passed across these runs. The focused tooltip/layout suite passed **142 tests**.
- Changed-file Ruff, formatting, and whitespace checks passed after rebasing; the earlier project-wide Ruff run used version **0.11.4**.
- All **nine targeted mutations** were caught: missing timer start, 100-second interval, one-shot timer, missing screen conversion, body coordinates passed to cell hit-testing, wrong popup parent, missing timer stop, missing stopped-state guard, and missing popup exclusion. This is a targeted check, not a rerun of the reviewer's full 66-mutant campaign.
- Acceptance tracing, an independent review starting from the original requirements, and final patch/commit review found no blocking defects. Final review also checked regression coverage, compatibility, commit messages, and exclusion of temporary artifacts.

The earlier pre-rebase native validation used labeled sample wx controls on macOS with wxPython 4.3.1 / wxWidgets 3.3.3 and the production controller/factory, not the running KiCad dialog. Native Windows/GTK, natural screen-edge flipping, and physical keyboard/wheel delivery remain untested. The extra non-Type check after main/table recreation used an invalid fixture coordinate and is not claimed as verified; both help types were verified after auxiliary recreation. The exact one-pixel dwell-reset behavior is unchanged; movement tolerance remains a separate optional improvement. Exploratory runs affected by activation setup or external pointer movement were excluded from the validation evidence.

The rebase-native fixture used wxPython 4.3.1 / wxWidgets 3.3.3 on macOS, with real controls, timers, activation, and popup rendering. It did not run the complete KiCad main-window constructor. Synthetic late timer callbacks after stop/destruction were safe. Physical-pointer trials interrupted by external pointer movement were excluded from the pass evidence. Native Windows/GTK and physical keyboard/wheel delivery remain untested.


z2amiller:
A couple of minor wx-isms:

* On macOS and Linux the hover code finds the window under the pointer without regard to which window is in front. Leaving Part Details or Why Standard open behind the plugin window kills both the fee popup and the existing Standard-only tooltip over that area. It's measured on macOS; Linux follows from the wx source.

* The timer and the geometry — the two genuinely wxWidgets parts — have no coverage. NEW. From a 66-mutant run, 47 caught and 19 survived, 11 of them real. Deleting self._timer.Start(100) entirely leaves all 30 controller tests green, as do changing the period to 100 s and making it one-shot, because the fixture fires the timer event by hand. Dropping ScreenToClient also survives, because the double's HitTest discards its argument and ScreenToClient is the identity — this is the one thing the deleted on_footprint_list_motion got right. And no test anywhere proves that the popup the controller shows is the popup containing the fee table: each of the two new test files replaces the other's module with a mock, so re-parenting the popup away from the list survives too. All of these are closable with assertions on doubles that already exist; none needs a display

* The 0.6 s dwell compares pointer position with exact tuple equality, so one pixel of movement restarts the wait, where wx's own tooltips carry a movement tolerance.  I manually tested this with a trackpad on my macbook and didn't have much trouble getting the popup to fire -- but other pointing devices may be less steady, not as an observed defect.

z2amiller (approved):
Things look good - there's a merge conflict now, probably due to my merging of 831 - if all of the conflicts are mechanical to fix, you're free to self-merge if you'd like.

PR #831, opened 2026-09-12
fix(ui): keep BOM and POS filters available without selection

chmorgan:
Hide excluded BOM and Hide excluded POS were disabled whenever no table row was selected. Keep both view filters enabled independently of selection so excluded rows can be shown again from an empty table.

Only the filters are removed from the selection-controlled toolbar group. The existing selection-event handling for footprint actions and the whole-toolbar disable for unavailable project storage are unchanged. A targeted Ruff exemption retains Python 3.9-compatible optional annotations.

The branch is rebased onto upstream `main` at `7dee28b`. The regression fixture intercepts `JLCPCBTools.Bind`, so it supports the main window's `wx.Frame` base. Every filter scenario runs with both notifying and silent model resets, with the fake selection cleared in either case.

Acceptance evidence:

| Behavior | Evidence |
| --- | --- |
| Filters work with zero, one, or multiple selected rows and after explicit deselection | Constructor and bound-handler regressions with stateful controls; native macOS Frame probe checks actual toolbar enable state. Explicit selection-event tests also verify footprint-action enablement. |
| BOM and POS filters work separately and together | Regressions cover both activation orders and verify visible rows, labels, and checked state. The native macOS probe exercises both filters and a combined sequence through dispatched toolbar events. |
| Hiding the last selected row still allows restoring it | Both filters remain enabled and restore rows whether model reset dispatches a selection event or stays silent. Native macOS controls also pass this workflow. |
| Filtering leaves exclusion values unchanged | Tests compare fixture BOM/POS exclusion flags before and after filtering. |
| Reopening with filters previously active starts with all rows visible | Constructor regression and native macOS probe verify the existing reset-on-open behavior. |
| Storage failure still disables the toolbar | Regression and native macOS checks inject a storage error; regression coverage also checks toolbar recovery. |

Validation:

- **1,932 tests passed:** `python -m pytest -q`. This includes **12 filter regression cases**, covering both reset-notification behaviors.
- The rebased branch reproduced the five reported fixture `KeyError` failures before the binding fix. The original filter regressions and native probe had reproduced selection-dependent disabling before the production fix.
- Ruff check, Ruff format check, and `git diff --check` passed. The feedback-fix commit passed Ruff, formatting, and pyupgrade hooks. The original production commit skipped pyupgrade's unrelated rewrites of pre-existing annotations.
- Native macOS verification passed with wxPython 4.3.1 / wxWidgets 3.3.3 using the actual Frame constructor, data model, table, and toolbar with fixture board/store data and dispatched wx events.

Platform limit and existing behavior: model reset does not guarantee a selection-change event. The [wx model documentation](https://wxpython.org/Phoenix/docs/html/wx.dataview.DataViewModel.html), [generic control source](https://github.com/wxWidgets/wxWidgets/blob/v3.3.1/src/generic/datavgen.cpp#L3143), and [GTK source](https://github.com/wxWidgets/wxWidgets/blob/v3.3.1/src/gtk/dataview.cpp) support testing silent resets. Footprint-action buttons can remain enabled after a silent reset until a later selection event; that pre-existing behavior is outside this filter fix. Automatic disabling after native model clearing was observed only in the tested macOS Cocoa run.

Acceptance tracing, independent requirements/platform review, and final patch review are complete. Windows/GTK behavior was source-inspected, not natively executed. Live KiCad integration, physical clicks, and real database persistence were not exercised. Temporary probe tooling is outside the repository.

### DIFF
--- a/mainwindow.py
+++ b/mainwindow.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 # pyright: reportMissingImports=false, reportMissingModuleSource=false
-# ruff: noqa: I001
+# ruff: noqa: I001, UP045
 
 from collections.abc import Iterable, Sequence
 from contextlib import contextmanager, suppress
@@ -1788,8 +1788,8 @@ class JLCPCBTools(wx.Frame):
             # cause pcbnew to refresh the board with the changes to the selected footprint(s)
             self.pcbnew.Refresh()
 
-    def enable_part_specific_toolbar_buttons(self, state):
-        """Control the state of all the buttons that relate to parts in toolbar on the right side."""
+    def enable_part_specific_toolbar_buttons(self, state: bool) -> None:
+        """Enable toolbar actions that operate on selected footprints."""
         for button in (
             ID_SELECT_PART,
             ID_REMOVE_LCSC_NUMBER,
@@ -1797,8 +1797,6 @@ class JLCPCBTools(wx.Frame):
             ID_TOGGLE_BOM,
             ID_TOGGLE_POS,
             ID_PART_DETAILS,
-            ID_HIDE_BOM,
-            ID_HIDE_POS,
         ):
             self.right_toolbar.EnableTool(button, state)
 
--- a/tests/test_mainwindow_filters.py
+++ b/tests/test_mainwindow_filters.py
@@ -48,11 +48,13 @@ def _toolbar(*_args: Any, **_kwargs: Any) -> MagicMock:
     return toolbar
 
 
-@pytest.fixture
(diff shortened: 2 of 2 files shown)


z2amiller:
Looks good, a few nitpicks:

* The new fixture patches test_window_layout._Dialog.Bind. Since #829 the window is a _Frame, so 5 of the 6 tests fail with KeyError on the merge. (One-line fix: patch mainwindow.JLCPCBTools.Bind instead. The full suite then passes, 1896 at head and 1926 merged)
* The deselect fake is macOS-only, and the claim that native model clearing disables row actions is false on Windows and GTK. (But I think this behavior already exists, it's not new, maybe worth digging into in a followup)

PR #833, opened 2026-09-14
Add native KiCad 10 design variant comparison and fabrication

chmorgan:
### CI integration fixes and Andy's review follow-up

Updated head: `aa0c54f96997c40adbe0f7fc6d17a88dbac4e99d`, based on main `c8005d2057062c9804cf1641cffc201d08959f2f`. All 29 feature commits are retained. This update changes four test files, with the repairs folded into their owning commits: pad metadata tests (`e022f64`), storage compatibility tests (`4276359`), and native matrix/dialog tests (`ddd0492`).

[CI on the preceding head `4a21009`](https://github.com/Bouni/kicad-jlcpcb-tools/actions/runs/35550079999) reported 22 headless failures and 24 native-controls failures (the same 22 plus two native cases). The failures exposed four test-integration issues after the rebase:

| Requirement | Fix and source-review evidence | Remaining evidence gap |
| --- | --- | --- |
| Run legacy CSV scenarios through the real electrical-pad metadata helper | Replace plain pad objects with SMD doubles exposing `GetAttribute()` and `HasHole()`. The existing pad count, through-hole, board priority, CSV preservation, reopening, and old SQLite schema assertions remain. This addresses the common constructor exception in 21 cases. | The previously blocked assertions have not been rerun locally. |
| Preserve main's retirement of automatic CSV import | Replace the obsolete import/backup expectation with empty and persisted SQLite-assignment cases. Both reopen twice, retain board exclusion flags and exact CSV bytes at the original path, and require no backup or migration metadata. Other schema and cross-mode checks remain. | Revised expectations have source review only. |
| Exercise native startup with unavailable settings writes | Inject the failure into the actual dialog instance passed to the real controller constructor. This avoids patching the temporary lazy-import wrapper and retains assertions for startup without writes and rollback after an explicit output change fails. | Native startup and event behavior have not been rerun locally. |
| Preserve translated output labels and canonical variant identity | Obtain the expected Default display label from the native board fixture, which deliberately supplies `< Par défaut >`. Empty canonical IDs, output switching, typed correction-row selection, LCSC edits, overlapping row IDs, and Default-value pattern fallback remain covered. | The old CI run stopped at the first label assertion; the later modal-edit and fallback steps still await runtime evidence. |
| Retain the reviewed history and scope | Acceptance tracing, an independent review starting from CI reports, and final patch/history review found no source-level blockers. The final tree matches the reviewed pre-fold tree; only the three owning commit patches changed. Commit messages were inspected; no temporary artifacts or agent guidance are included. | Source and history review do not establish runtime correctness. |

**Validation status:** Per the requested scope, no tests, lint, formatting checks, builds, native UI/KiCad execution, or other runtime validation were run locally for this combined revision. Updated-head CI results are pending. Native KiCad, isolated-input, and lint jobs passed on `4a21009`; those results belong to that earlier head.

Andy's requested conditional `msvcrt`/`fcntl` imports remain at module scope, with OS-specific lock/unlock operations retained. The rebase retains main's KiCad 7 minimum, lazy startup import and error reporting, CSV retirement, Python 3.9 settings, and per-LCSC corrections. Output-specific LCSC correction resolution is shared by display, editing, and fabrication; pattern fallback remains anchored to Default reference/Value/footprint. Windows locking remains unexecuted locally.

The prior acceptance and validation evidence below applies only to its explicitly named earlier heads.

---

Boards with named KiCad 10 design variants now open a side-by-side table for comparison and editing, and fabrication uses an explicitly selected output variant. Boards without named variants retain the ordinary parts table.

The table supports per-variant Value, LCSC and BOM/POS/POP edits, batch assignment and copying, difference highlights and filtering, price/stock details, and dragging complete variant column groups with a visible preview. Native board fields remain the source of truth; saving the PCB preserves edits. Variant modules are included in PCM packaging, and the README includes three animated workflow examples.

The shared workflows now handle BOM estimates and electrical-pad counts, modal-window cleanup, overlapping parts lookups, catalog recovery, concurrent window preferences, and publication of fabrication files with their generation count.

### Fixes for the failing KiCad job

The original failures were reproduced with the real KiCad 10.0.6 package in Docker: 4 failed and 13 passed. Temporary board save/load now performs file access in Python, verifies a digest calculated before writing, and gives KiCad the exact verified bytes in memory. Directory, missing-file and corrupt-file failures therefore reach Python cleanup instead of aborting inside untranslated C++ exceptions.

The project variable was present: the failing test indexed native `wxString` keys with a Python string. Assertions now normalize the keys and values while still requiring the expected project variable. Investigation also found two native problems: promoting borrowed SWIG wrappers to owners eventually invalidated the wrapper registry, and the full save formatter changed embedded fonts on the source board. Explicit one-time cleanup and a disposable formatter context address those problems. Test doubles restore their imports, and native tests reuse the installed KiCad module.

These changes retain and extend the real KiCad tests. The earlier local headless run had not executed that selection. The README now documents all four test selections and reproducing the Linux jobs with Docker.

### Acceptance evidence

| Requirement | Evidence |
| --- | --- |
| Preserve native variant state, inheritance and existing storage across reopening | Session, adapter, storage, matrix and export tests cover named/default variants, removed variants, stale boards and existing metadata. Real KiCad tests also check distinct variables and project identity in two fresh, initialized and saved/reopened projects. |
| Recover from temporary board save/load failures without changing the source | Native directory, read-only, missing-file, corruption and publication-validation cases raise Python exceptions, remove temporary directories and successfully retry in the same session. Unsaved text reaches the copy; source PCB/project bytes and selected source variant stay unchanged. |
| Release copies safely and preserve board content | Twenty native copy/release cycles check collection and finalizer completion. Four native combinations cover two/six copper layers, pending/retained embedded fonts, custom layer names, pad layers, nested group membership, failed writes and complete serialization roundtrips. Restoring the old formatter makes all four font regressions fail. |
| Reject changed bytes before native parsing | Boundary tests corrupt bytes during writing and replace the path after reading. The digest rejects corruption, and the parser receives the immutable bytes already verified. |
| Preserve matrix interaction and rendering | Native wx controls cover selection, filtering, column order/preview, menus, modeless dialogs and window lifecycle. Separate mouse/keyboard tests run with an isolated display and window manager. GTK regressions cover opaque drag previews and highlights after ligatures. |
| Recover assignments and generation after catalog failure | Native-window regressions exercise repeated healthy retries, actual selector events, native-adapter assignments and real ZIP/CSV generation; physical plotting is simulated in these recovery tests. Replaced boards remain disabled. |
| Preserve independent preferences when windows save or close | Process contention and 12 native simultaneous-window cases cover different boards, independent fields, ordinary-window saves, stale controls, absent defaults and reopening. Failed writes preserve disk and memory. |
| Keep outputs and generation count consistent on recoverable failures | SQLite and handler tests cover read-only databases, competing writers, stale counts, COMMIT failure, restoration failure, publication-lock contention, cleanup, successful retry, one increment and post-hook ordering. New handler cases cover initial and publication-time board checks; native tests separately exercise actual KiCad save/load. |
| Preserve ordinary workflows | Full headless and native wx suites cover BOM/pricing, pads, corrections, lookup retries, dialog ownership, settings migration and ordinary-board exports. Existing real KiCad SVG/Gerber assertions also pass. |

### Rebase onto main after #834

Rebased onto `1374457` (merged #834). Final head: `90618e33795b2d13c0a6271166063869eee1e8dd`. The 29 feature commits are retained, with the resolution and regression tests folded into the existing lookup commits.

The ordinary table keeps #834's Std indicators and assembly tooltips while using the shared lookup scheduler. Request, result and completion callbacks refresh row help; results project merged cached metadata immediately, and completion preserves loading state owned by other batches. Reconciliation changes are limited to `mainwindow.py`, `tests/test_mainwindow_enrichment.py` and `tests/test_mainwindow_assembly_lookup.py`.

| Rebase requirement | Evidence on the rebased tree |
| --- | --- |
| Preserve Std classification, cached process/classification, pending state and hover refresh | All upstream enrichment scenarios use production scheduler callbacks with queued delivery and real model/SQLite state. Empty results, partial failure, cached classification and reopening are covered. Before the fix, six adapted cases failed for missing metadata/hover updates. |
| Preserve overlapping assignments, late subscribers and storage recovery | Native macOS window tests exercise independent/shared requests, reassignment, completion while another code is pending, storage replacement/reopening and thread-start failure/retry. The new native assertion first reproduced an Economic result displaying `?`; it now reads the expected classification from the model attached to the real control. |
| Preserve named-variant lifecycle and catalog recovery | The focused native selection includes variant readiness, main-window lifecycle and catalog recovery: 57 passed with no skips. |
| Preserve table layout and saved settings | Indicator, tooltip and layout tests retain hidden Enrichment, inert text Std, stable model indices, metadata reset and legacy column-width reopening assertions. The complete headless selection passed: 2,470 passed, 3 existing opt-in legacy skips, 433 native cases deselected. |
| Keep scope and history intact | Acceptance tracing, independent requirements/implementation review and final range-diff/commit review completed. The final folded tree exactly matches the validated tree; main is an ancestor and whitespace checks pass. |

Ruff 0.11.4 lint and formatting checks passed for all three reconciliation files. `pre-commit` and `pyupgrade` are unavailable in the configured environment, so those hooks were not run locally. CI for this head is pending. Native validation used wx 4.3.1 on macOS with real controls/event delivery and deterministic board/supplier boundaries; rendered glyph pixels and the visible hover popup were not inspected. Linux, Windows and real KiCad lanes were not rerun locally for this rebase. Earlier validation below applies to the explicitly named previous head.

### Validation before this rebase

**All seven GitHub CI checks pass on `2f17364547dd6ca030ed7185ee1e2a2a8dd8621c`.** The [test workflow](https://github.com/Bouni/kicad-jlcpcb-tools/actions/runs/34892667291) passed with 2,426 headless tests, 2,831 native wx controls tests, 2 isolated input tests, and [26 real KiCad tests](https://github.com/Bouni/kicad-jlcpcb-tools/actions/runs/34892667291/job/104138920608). [Markdownlint, Ruff and Deadcode](https://github.com/Bouni/kicad-jlcpcb-tools/actions/runs/34892667325) also pass. GitHub installed the same KiCad 10.0.6 package used locally; no required native test skipped.

The final tracked tree was validated locally with packaged KiCad `10.0.6~ubuntu24.04.1` in Ubuntu 24.04 Docker, using a non-root runner, the workflow selections, strict native requirements, Xvfb, timeouts and Openbox for input tests. Docker contains KiCad and wx; the separate CI wx job installs wx alone.

| Check | Result |
| --- | --- |
| Local headless | 2,426 passed; 3 skipped |
| Docker native wx controls, including headless | 2,831 passed; 3 skipped |
| Docker isolated mouse/keyboard | 2 passed; zero skips |
| Docker real KiCad | 26 passed; zero skips |
| Ruff 0.11.4 | All 185 non-ignored tracked Python files passed |
| Formatting and Markdown | Six changed Python files and README passed |
| Deadcode 2.4.1 | Exit 0; 136 advisory findings, unchanged from baseline |
| Whitespace | Passed |

The three skips are existing opt-in legacy desktop tests in `test_why_standard_dialog.py`; no required native test skipped. Those legacy tests were previously exercised separately on macOS. The final schematic warning-level adjustment also passed its 57-test host suite.

Acceptance tracing, an independent review starting from the reported failures, and final patch review led to the font, ownership, retry and project-isolation regressions above. The fixes are folded into their owning commits, with redundant repair commits removed; the rewritten final tree is identical to the validated tree.

### Remaining limits

The native execution covers Linux KiCad 10.0.6. Other KiCad versions, native KiCad on macOS/Windows, and a live PCB Editor's undo/dirty state and drawing-sheet/layer globals were not exercised. Avoiding public helper side effects is supported by KiCad source inspection. The formatter context depends on KiCad implementation behavior, and the parser assumes KiCad can read its own unchanged formatted output.

Windows locking has not been executed locally. File/database compensation handles caught runtime failures; it does not promise atomic recovery across process termination or power loss.


core/file_lock.py
```diff
@@ -0,0 +1,88 @@
...
+from typing import BinaryIO, Optional, Union
+
+_LOCKS: dict[str, Lock] = {}
+_REGISTRY_LOCK = Lock()
+
+
+def _try_file_lock(handle: BinaryIO) -> None:
+    """Acquire an OS lock without blocking beyond the caller's deadline."""
+    if os.name == "nt":
+        import msvcrt
```
z2amiller, line 21: Could you move the conditional imports up into the main import block?
(You'd still need to check the OS  here, of course, but would only trigger
the imports once)

z2amiller (approved):
This looks good, aside from needing to be merged (which looks pretty clean)
plus the nitpick about imports.

chmorgan:
@z2amiller, could you re-review the updated PR at `4a21009`?

The conditional imports are now at module scope, and the branch is rebased onto current main. The update also preserves the newly merged per-LCSC corrections across variant display, editing, and fabrication. Fixes are folded into their original commits.

This combined revision has had source and history review only; tests, lint, builds, and native checks were not rerun. The PR description separates the current review evidence from earlier validation results.

z2amiller:
Re-reviewed 4a21009 by merging the previously approved head (90618e3) with current main and diffing that against this head, so only the manual resolution was in view. The resolution looks right to me: exact LCSC rules follow the output variant's assigned part, pattern rules still resolve from Default, and the matrix column, hover text, correction editor and CPL all go through the same resolve_shared_corrections call. A deleted output variant reads "Unavailable" rather than borrowing Default's angle. The import change is what I asked for, and plugin.py keeps main's version gate ahead of the constructor guard. Range-diff shows all 29 commits retained.

The two red lanes reproduce locally (macOS, Python 3.12, wxPython 4.3.1) with CI's exact counts: 22 headless, 24 native. None are timeouts. They come from four test-side rebase interactions:

* tests/test_store_legacy_csv.py, 21 cases. Main's #836 doubles return bare object() pads, and this branch's stricter count_pad calls GetAttribute() on them. These tests assert the legacy path is ignored, so keep them; the doubles just need GetAttribute() and HasHole().
* tests/test_store_main_compatibility.py::test_ordinary_csv_import_keeps_backup_behavior_without_new_marker. Asserts the CSV import that main removed. Delete it; we want that path gone.
* tests/test_variant_controller_corrections.py:130. Expects Output variant: Default, but the native fixture labels Default < Par défaut > on purpose. Use model.variant_label(variant).
* tests/test_variant_controller.py:665. Patches save_settings on mainwindow.JLCPCBTools, but run_frames now replaces that name with the constructor function since plugin.py imports it lazily. Patch type(args[0]) instead.
The attached patch makes those four changes. With it applied: headless 2647 passed, native controls 3053 passed, ruff 0.11.4 clean.

One non-blocking note: when a named variant's LCSC is blank, pattern fallback uses Default's Value rather than the variant's. That is consistent with the shared-rule design and the new pr-833-comment.md
test covers it deliberately, but a sentence in the README would save a user who writes Value rules per variant some confusion.

Approval stands once CI is green.

Patch to get it working:
[pr-833-test-fixes.patch](https://github.com/user-attachments/files/32448330/pr-833-test-fixes.patch)

chmorgan:
@z2amiller, thanks for the re-review and patch. The latest push, `aa0c54f`, should address the four test issues you identified, with the fixes folded into their owning commits. All seven CI checks are now passing.

I had asked for your re-review before seeing CI pass, so the review request got ahead of that confirmation. The README also already notes that pattern rules use Default's reference, Value, and placed footprint.

z2amiller (approved):

PR #843, opened 2026-09-16
Add bulk LCSC assignment for matching value and footprint

RolandWa:
> **Note:** This PR is one of five focused PRs split from #784 at the reviewer's request to allow each feature to be considered independently.

## Summary

Adds a right-click context menu action **Assign LCSC to matching value and footprint** that propagates the selected part's LCSC number to every other component with the same value and footprint in one step.

### Behaviour

1. User selects a component that already has an LCSC number.
2. User right-clicks and chooses **Assign LCSC to matching value and footprint** (placed at the top of the context menu).
3. A confirmation dialog lists the affected references (up to 10 shown, then "...").
4. On confirmation, all matching components are updated in:
   - The project store (SQLite)
   - The KiCad board footprints
   - The data model / UI
   - Enrichment / assembly data (async)
   - The BOM estimate

The action is a no-op if the selected part has no LCSC number, or if no other part with the same value and footprint needs updating.

## Validation

- `python -m compileall -q mainwindow.py`
- `git diff --check`

## Manual Test Plan

1. Open a project with several resistors/capacitors of the same value and footprint.
2. Assign an LCSC number to one of them.
3. Right-click it and choose **Assign LCSC to matching value and footprint**.
4. Confirm the dialog lists the expected references.
5. Confirm all listed parts are updated after accepting.


z2amiller:
This already exists in the UI "Auto-assign alike footprints" on the right hand side.  If it needs to be duplicated in the right-click menu, can it at least share the existing code for doing that?  It looks like it was re-implemented.

Issue #835, opened 2026-09-15
Remove part_info and use KiCad board data instead


chmorgan:
# Remove part_info table

## Purpose

There are issues when opening more than one board in the same directory and using this plugin. Much of this stems from the fact that we store per-part reference information in the part_info table. If you assign say R1 to one part in one board, then open another board in the same folder the plug-in can map R1 to the wrong component.

Almost all of the entries in this table are duplicates with the data kept in kicad itself, the LCSC number, whether its in the BOM etc.

One approach is to add per-board tracking and identification to this table and to the code.

This approach is the opposite. It removes ALL overlapping information from part_info and defer to the kicad data.

While this isn't a large net savings of lines here, we can eventually remove the migration code and see some savings, plus less duplicate data is a good thing.

For the LCSC number specific data this data is moved to the global mapping.db, things like stock numbers etc that are pulled at run-time and that we want to cache. Previously these were in the part_info table, mapping.db seemed like a good place for them.

## Plan

1. **Keep and simplify `Store`**, preserving its read API instead of introducing separate board-reader and project-metadata services.
2. **Use a three-column global cache**, preserving the current fetch-missing behavior.
3. **Put migration in one removable file**, using standard dialogs and source-file/table presence to determine completion.

This plan is based on **[latest `origin/main`, `e7884e0`](https://github.com/chmorgan/kicad-jlcpcb-tools/commit/e7884e0df61efb59931c13b7949658e96c944e55)**, reverified through GitHub. It uses no working-branch changes.

**Revised estimate:** approximately **235 fewer permanent runtime lines**. Including the temporary migration implementation, the first release would be approximately **110 lines larger overall**.

### Project database: `<PCB directory>/jlcpcb/project.db`

**Remove:**

```sql
DROP TABLE part_info;
```

Do this only after successful migration.

This removes persisted copies of:

- Reference, value, footprint, and LCSC assignment.
- BOM/POS exclusions.
- Stock.
- Pad count, through-hole state, and assembly flags.
- Per-reference assembly-process and product-classification metadata.

**Retain:**

- `metadata`, including `generation_count`.
- Existing project correction tables, including `correction`, `correction_migrations`, and `correction_migration_state`, wherever present.
- Any unrelated tables.

**Add:** no replacement parts table, board-identity table, migration table, or migration-completion key.

The existing `metadata` table can be initialized lazily when generation-counter persistence is needed.

### Global database: `<Library.datadir>/mappings.db`

[...]

PR #845, opened 2026-09-16
Major - Workflow changes to data management

chmorgan:
This is the best place to start when reviewing this PR: https://github.com/chmorgan/kicad-jlcpcb-tools/tree/feature/plugin-owned-assembly-data#data-flow-and-ownership

The stuff below here is mostly AI slop and doesn't really get to the core of the change like the README.md section does.

## Problem and result

Updating a PCB from its schematic can replace LCSC assignments and assembly flags. Reference-only project records also let boards in the same directory share the wrong assignments (#835).

This change makes **kicad-jlcpcb-tools authoritative for LCSC, BOM/POS exclusions, and DNP**. A missing record imports PCB values once. Existing records, including deliberately blank assignments, survive schematic updates. Plugin edits save immediately to SQLite and never write native PCB assembly settings or schematic files.

Started from latest upstream main `b1ce6ed` in a new feature branch. This replaces the direction explored in #839; that PR and its branch are unchanged. No review requested yet.

## Database changes

No new database file.

- **Existing `jlcpcb/project.db`:** active `part_info` now contains `board_key`, `footprint_uuid`, `lcsc`, `exclude_from_bom`, `exclude_from_pos`, and `is_dnp`. Its primary key is `(board_key, footprint_uuid)`; flags are constrained to 0/1.
- Remove persisted duplicates of reference, value, footprint, stock, pad/THT geometry, serialized assembly flags, and supplier classification from the active table. Resolve design facts from the current PCB and supplier facts by the stored LCSC code.
- **`part_info_migration.py`:** transactionally replace the old table, converting matching references to UUIDs and discarding unmatched legacy rows. Remove any existing `part_info_legacy` table and its `part_info_legacy_owner` bookkeeping. Ambiguous board ownership or changed value/package requires confirmation; cancellation or failure retains the original database. Keep the existing project `metadata` table for generation counts and board-scoped notice acknowledgment.
- **Existing global `mappings.db`:** add `lcsc_metadata(lcsc TEXT PRIMARY KEY NOT NULL, assembly_process TEXT, component_product_type INTEGER)`, with nullable product type constrained to 0/1/2. Preserve the existing preference table. No timestamps, expiry service, or new catalog schema.
- Keep correction storage. Stock, prices, and descriptions remain catalog data. Do not restore legacy CSV import.

A footprint rename preserves its UUID assignment; replacement means a new part. Deleted records remain available for undo, while current outputs contain only live footprints. The board key is relative to the project directory, so moving the complete project preserves records. A renamed/copied board path starts separate choices; this PR adds no implicit cross-board transfer.

## Implementation and simplification

- Replace overwrite/reconciliation branches and geometry backfilling with import-if-missing plus one resolved Store view.
- Use atomic, UUID-validated database edits for assignment, clear, BOM/POS, and the new DNP context action. Reject stale selected batches without partial success.
- Automatic preferences apply only to eligible new records. Successful explicit edits consume that eligibility, including when catalog availability is deferred.
- All table, stock, estimate, BOM and CPL consumers use authoritative assembly choices. Capture one manufacturing snapshot and reject changed membership before output writes; physical placement remains live.
- Merge each supplier response once per LCSC code and refresh matching rows from one snapshot. Cache failures remain independent of authoritative assignment commits.
- Remove Save/Export to schematic, its dedicated assets/tests, the priority setting, native assembly mutation helpers, and unused model togglers.
- Detect saved schematic LCSC fields through a matching KiCad CLI's generic XML netlist export. Run asynchronously with a timeout; acknowledge the advisory once per board. Recommend cleanup in Schematic Editor; never modify its files.
- README: explain the historical synchronization problems, document present and future data ownership with separate Mermaid diagrams and an ownership table, and distinguish PCB design attributes from plugin assembly choices. Both PCB boxes list part value and footprint; diagram labels use consistent font sizes and keep generated-file details on separate, unwrapped lines.

## Validation and acceptance evidence

| Requirement | Evidence |
| --- | --- |
| Existing choices survive schematic updates | Fail-first real SQLite regressions cover removed/blank/stale LCSC, changed flags, and reopening. Native KiCad 10.0.6 workflow below confirms stale schematic LCSC does not replace plugin assignments. |
| No native assembly writes | Real picker and flag/DNP actions left native fields/attributes unchanged; controller tests cover assignment, clear, preferences, and atomic failures. |
| Identity and preservation | Store tests cover two boards with overlapping references/UUIDs, reference rename, replacement UUID, removal/undo, project-folder move, legacy ambiguity, cancellation and transactional DDL rollback. Successful migration retains matching assignments without an archive and discards unmatched rows. Tests also cover cleanup of an existing archive and rollback of that cleanup when schema validation fails. |
| Intentional blanks | Clear/reopen and deferred-catalog regressions pass, including an initially eligible blank explicitly cleared before preferences can run. Native reopening also retained a deliberate blank while PCB LCSC remained populated. |
| Consistent manufacturing data | Conflicting native/DB DNP and exclusion tests, BOM grouping/chunking, snapshots, and membership/UUID validation pass. Native BOM/CPL generated through the real Fabrication object agree with stored choices. |
| Supplier cache | Tests cover partial merges, class zero, separate library instances, failed writes, path isolation, recovery, and large batches. |
| Notice and lifecycle | Tests cover unknown/error outcomes, acknowledgement, cancellation, a second window, database failure, stale callbacks, closed windows and Save As. Actual native notice appeared once; the real CLI detected assigned/excluded/DNP symbols without changing PCB/project/schematic hashes. |
| Removed behavior and compatibility | Constructor/settings tests verify export/priority controls are gone; standalone footprints have stable UUIDs. Ruff, formatting, pyupgrade, and Markdown checks passed as detailed below. |

### Native macOS KiCad 10.0.6

Used a disposable schematic-linked resistor project and isolated plugin settings/catalog storage.

1. Opened the actual plugin and acknowledged the detected schematic LCSC notice.
2. Assigned `C25804` through the actual part picker; toggled BOM/POS and DNP through native controls. Database choices changed while PCB `C300`, attributes `2`, and DNP false remained unchanged.
3. Generated excluded BOM/CPL through the real Fabrication methods: only headers remained.
4. Edited the schematic LCSC in KiCad Symbol Properties, then executed actual Update PCB from Schematic with field updates enabled. The resulting saved PCB contained `C300`; the plugin reopened showing `C25804` and retained its exclusions, without repeating the notice.
5. Re-enabled inclusion in the plugin. Actual generated BOM contained `10k,R1,R_0603_1608Metric,C25804,1`; CPL contained the current physical placement.
6. Cleared through the real handler and reopened: plugin assignment remained blank while native PCB LCSC remained `C300`.

The native BOM/CPL checks invoked the real production methods in KiCad's console after real UI edits. They did not run a fresh full Gerber/DRC generation. Automated tests cover generation preflight and hook ordering.

### Checks and reviews

- Earlier full-suite run, before the latest archive-removal changes: **2,073 passed, 3 skipped** in an isolated tracked-file snapshot.
- Latest archive-removal update: **137 focused tests passed**, using real Store construction and SQLite with fake PCB objects. The new archive-removal expectations failed against the previous implementation. Coverage includes matching assignment preservation, unmatched-row removal, cancellation, rollback after a partial conversion, existing-archive cleanup, and cleanup rollback on schema-validation failure. Repository virtual-environment Ruff and formatting checks passed; an independent review found no gaps.
- Latest README changes passed the Markdown hook; the new font sizes and no-wrap styling have not been visually verified in GitHub.
- Python changes passed repository-environment Ruff and formatting checks; pyupgrade and applicable Markdown hooks passed. An unrelated annotation rewrite from the pinned Ruff hook was excluded; repository hook configuration is unchanged.
- Acceptance trace, independent requirements review, and final patch review completed. Findings received fixes and regressions: stale partial batches, deferred preferences, repeated enrichment scans, duplicate notice delivery, standalone UUIDs, and snapshot membership changes.

## Size against upstream main

| Area | Added | Removed | Net |
| --- | ---: | ---: | ---: |
| Production | 1,097 | 1,167 | -70 |
| Tests and fixtures | 1,359 | 2,193 | -834 |
| README | 126 | 7 | +119 |
| **Total** | **2,582** | **3,367** | **-785** |

Production plus tests remove **904 net lines**, exceeding the 500-line target. Removed coverage belongs to retired exporter/CSV/reconciliation behavior; replacement tests cover the new ownership contract.

## Remaining limits

Native verification used macOS KiCad 10.0.6 and a small board. Native KiCad 7–9 and Windows/Linux were not exercised. The notice reads saved effective netlist fields, not unsaved editor changes or private fields omitted by KiCad; unidentified/differently named roots or unavailable CLI return unknown and do not consume acknowledgment. Complete supplier metadata retains the existing fetch-missing policy. Successful migration discards unmatched legacy rows and removes legacy archives; automatic downgrade or board-renaming transfer is outside this change.


chmorgan:
@z2amiller alright, this is the alternative approach I mentioned working on.

tldr: We stop synchronizing any schematic driven data (any schematic fields) back to kicad entirely. It isn't and hasn't been safe to be saving schematic files manually, and we could overwrite open files, there is no api that lets us manipulate the schematic, and as soon as the user hits 'update pcb from schematic' they'll overwrite the LCSC values from the schematic.

This PR removes the ability to write back to the schematic file. I should have spotted this earlier but this is a bad idea and imo there aren't good reasons to do that when the workflow of the user updating the schematic to get jlcpcb updates is effectively broken at present due to kicad limitations.

This isn't an ideal set of changes, however it does simplify and resolve a bunch of data flow issues and I'd claim its about the best we can do until kicad fixes their implementation, something we should lean on them to do. We should let them know the issues we see here and confirm they are aware of the limitations we are hitting.

Oh, and we still preserve the initial data import from kicad settings, its just that we only do that if a part isn't present in part_info.

Thoughts?

chmorgan:
@z2amiller and these diagrams and writeup (still draft) I think explain the change well visually. https://github.com/chmorgan/kicad-jlcpcb-tools/tree/feature/plugin-owned-assembly-data#data-flow-and-ownership

chmorgan:
@Bouni @z2amiller requesting another review of these changes. Imo the most approachable entry are these readme updates in the PR branch: https://github.com/chmorgan/kicad-jlcpcb-tools/tree/feature/plugin-owned-assembly-data#data-flow-and-ownership

I know it's a shift from what we do. I still think its the best we can do given the plugin limitations through kicad 10.

Bouni:
I'm on vacation right now and reviewing stuff on my phone sucks 😅
Feel free to merge if @z2amiller approves

chmorgan:
@Bouni this is defiantly a *we need to discuss and agree as a team* kind of change as we basically stop pushing changes back to the kicad design since we can't do it in a way that works consistently and reliably. You can probably view most of it on web if you click on that link in the comment. That's the readme updates related to this change.

Bouni:
I'm back on Monday and will try to find the time to look at it properly on a real screen

Bouni:
I tried to understand what this quite huge set of changes do. Getting rid of writing back to the schematic is ok for me as its a very hacky solution anyway. 
I think in the upcoming releases of KiCAD the IPC API will allow us to write back to the schematic in a safe way. To me the most important feature would be writing a chosen LCSC number back to the schematic symbol.
As well as setting DNP, and so on.

Has anyone of you guys looked into what the IPC API allows at the current state?
https://docs.kicad.org/kicad-python/schematic.html#kipy.schematic_types.SchematicField

What do you think of a kind of wizard window for reading from the schematic that clearly shows whats the state of the PCB/project.db and what would be read from the schematic?
That way we could allow a user to update from the schematic and select fine grained what should be updated!?

z2amiller:
@Bouni -- this is kind of what we came up with in the issue discussion.
(https://github.com/Bouni/kicad-jlcpcb-tools/issues/835)

We can be the source of truth for the fabrication attributes, but provide visibility when the schematic/PCB are out of sync with that.  (Which is especially important for BOM/POS/DNP since that will surprise users if they change that in the project and it doesn't change the fabrication). 

I think we'll probably need to maintain the PCB/schematic writeback for the time being, despite the hacky-ness.  Even if the KiCad team provides schema manipulation tools in KiCad 11 (which isn't for sure - AFAICT the Schema-related IPC methods are read only for now), we still have to support old versions for however long.

PR #857, opened 2026-09-22
Add controlled-impedance review and fabrication reports

chmorgan:
Boards requesting controlled impedance need JLCPCB's completed workbook and marked trace images alongside their Gerbers. This adds a board-specific configuration and review workflow, then includes `Required_impedance_control.xlsx` and a self-contained HTML companion in the Gerber ZIP when enabled.

Specifications select native net classes, impedance type and target, per-layer references and applicable gaps. The dialog provides capture review, approval history, JLCPCB stackup selection and nominal width comparisons. Saved drafts remain editable; export requires current approval. The reports record the designer's requirements and do not certify electrical impedance.

The workflow supports the explicit output variants now on `main`: report captures use the fabrication board copy, while net-class metadata comes from the guarded live source. Existing part assignments and generation counters retain their current storage model. Opening the plugin restores impedance settings read-only; tables and board registration are deferred until settings or downloaded catalog data are saved.

### Review guide

- `impedance/`: configuration, SQLite persistence, matching, capture/review, stackup tools and report writers.
- `mainwindow.py`, `fabrication.py`, `fabrication_archive.py`: toolbar integration, shared copper preparation, report validation and publication.
- `docs/` and `examples/impedance/`: user documentation, combined RF example boards and reproducible fixture/capture tooling.
- Much of the diff is the supplied workbook template, fixtures and bundled `openpyxl`, `et-xmlfile` and `websockets` sources with their license metadata. Lint exclusions preserve third-party sources and the copied KiCad license text.

### Behavior and regression evidence

| Requirement | Evidence |
| --- | --- |
| Enabled exports include both reports; disabled regeneration omits stale reports; failures preserve prior output | Actual XLSX/HTML/ZIP assertions in `test_impedance_report_pipeline.py`, archive tests, and production handler sequencing in `test_mainwindow_impedance_flow.py` |
| Cancel leaves saved state unchanged; drafts save/reopen without acquiring approval; failed saves retain edits | Stateful control/dialog tests in `test_impedance_controls.py` and `test_impedance_draft_dialog.py`, plus actual SQLite reopening in `test_impedance_draft_persistence.py` |
| Boards remain isolated across reopen/project moves; concurrent saves and schema upgrades preserve existing records | SQLite transaction, migration and exact-record checks in `test_impedance_database.py`, `test_impedance_schema5.py` and `test_store_schema_compatibility.py` |
| Changes to geometry, rendered text, project definitions or settings invalidate approval | Matching, context and tracking checks in `test_impedance_context_fidelity.py`, `test_impedance_project_context.py`, `test_impedance_text_context.py` and review-dialog tests |
| Review honors zone preparation; failed fills stop export; previews do not mutate source geometry | State-changing KiCad doubles in `test_fabrication_zone_fills.py`, `test_impedance_board_copy.py` and review-control tests |
| Output-variant mismatch blocks publication; live net classes do not replace the copy's geometry/text/pairing | Real archive/counter lifecycle with native boundaries doubled in `test_variant_generation_lifecycle.py`; adapter checks in `test_impedance_netclass_native.py` |
| Reports share immutable captures, preserve template fields and escape untrusted labels | Real workbook/HTML generation in `test_impedance_workbook.py`, `test_impedance_report_pipeline.py` and service tests |
| Variant startup/editing/reopening preserves absent, legacy and invalid project storage; saved impedance still restores | Real wx constructors, events and byte-preservation checks in `test_plugin_board_startup.py` and `test_variant_render_snapshot.py`; SQLite authorizer rejects non-feature reads and only URI `mode=ro` connections are permitted |
| First use remains optional; first catalog save persists without a config revision; concurrent first enable blocks stale export/save | Regression cases in `test_impedance_controls.py` and `test_impedance_database.py`, using real SQLite, exact saved approval fields and production dialog callbacks; configuration dialogs/board geometry are doubled in these cases |
| Native parameter highlights retain shaped glyph layout and exclude ellipsis under upstream package recognition | Canonical `R_0603_1608Metric` and abbreviated `R_0603` cases in `test_variant_matrix_layout.py`, with original clipping assertions retained |
| Distribution declarations include the feature and expected workbook resource | Package declarations and template hash checked by `test_impedance_packaging.py`; this is not an installed-package test |

### Validation

- All seven GitHub checks passed on `83ba4b3`: standard tests, native wx controls, isolated native wx input, native KiCad 10, Ruff, markdownlint and deadcode. [Test run](https://github.com/Bouni/kicad-jlcpcb-tools/actions/runs/35719441259), [lint run](https://github.com/Bouni/kicad-jlcpcb-tools/actions/runs/35719442782).
- Full tracked test suite with `not native_wx and not native_kicad`: **5,469 passed, 239 skipped, 437 deselected**.
- Affected native wx startup, variant events/reopening, preferences and layout checks: **111 passed**. Separate native layout/rendering and package-highlight checks: **148 passed**. These use real macOS wx controls with KiCad board doubles. Earlier targeted fabrication-window/post-hook checks also passed (2 tests).
- Ruff lint passes for all 289 tracked first-party Python files; formatting passes for all eight Python files in the CI fix (the original feature files were checked before the initial push).
- Markdown lint passes for the changed documentation; the verbatim KiCad license is excluded. Shell syntax and first-party diff whitespace checks pass.

### Remaining validation gaps

The local environment has no `pcbnew`, so this run does not establish native KiCad rendering, clone equivalence, SWIG ownership, net-class/pair bindings, zone filling or the complete impedance-dialog lifecycle. Real Excel/Numbers appearance, live JLCPCB solver behavior, Linux Docker execution, and built/installed package loading were not exercised. The green Linux CI native jobs cover the existing selected wx and KiCad workflows; they do not establish complete interactive impedance-dialog, image-rendering or electrical-validation coverage.


z2amiller:
Is this its own entire plugin?  How much of the bouni plugin does it really use?   It seems like it has
its own UI, own database(tables), etc.

Could this be a separate plugin that inspects the gerbers/etc?  I'm not sure how general purpose it is.
(Good excuse to learn IPC plugins :-) )

I guess I don't work in RF stuff so I don't really understand the domain or how it's used - are there
screenshots/mockups as to how this would integrate into the plugin?

z2amiller:
Some more early feedback:

- Could this be split out?  I think the stackup picker / JLC impedance calculator calls (and the associated websockets deps) could probably be split into their own PR - they're not core to the .xls writing functionality.
- For the websockets call to the solver, it fakes its origin as jlcpcb.com -- I dunno if that's something that they'd get irritated about.  Could it link out to a pre-filled form or something rather than be embedded in a UI in the plugin?
- Can you make sure that the PCM bundling is correct?  For example, we'll want to arrange things so that all of the new stuff in scripts/ doesn't become part of the overall package.  (Nor the docker configs, etc)
- Is it possible to drive more of the process thru the KiCad Dockerfile? ("Remaining validation gaps" -- it sounds like this isn't being driven in the tests)

PR #858, opened 2026-09-22
fix(search): read a bare chip size after its class letter

z2amiller:
Upstream `main` at 00b5da7 (the #855 merge) fails the **Native wx (controls)** lane: `tests/test_variant_matrix_layout.py::test_parameter_colors_keep_one_native_glyph_layout_and_exclude_ellipsis[False|True]` paints one coloured run where the test expects two. The lane passed on e445fed seconds earlier, so it is the change, not the runner.

## Cause

#855 made the chip-size rule require the imperial code together with its metric companion, `R_0603_1608Metric`. The variant matrix fixtures spell their default resistor `Resistor_SMD:R_0603`, so `simplify_footprint_name` now returns nothing for it, `expand_footprint` contributes no term, and the params cell only gets the value highlight (`10kΩ`), not the footprint one (`0603`).

## Which side is wrong

The bare spelling is a real input, not a fixture shorthand: KiCad 4's libraries wrote it that way (`Resistors_SMD:R_0603`, `R_0603_HandSoldering`; see the archived `KiCad/Resistors_SMD.pretty`), and a board built from them keeps those footprint ids through every upgrade. Before #855 those boards got `0603` prefilled and highlighted; after it they got nothing. So the rule is what regressed, and the upstream test is left as it was.

## Fix

A second pattern reads the code on its own, but only as a whole segment introduced by one of the classes the library pairs with a chip size: `R`, `C`, `L`, `D`, `LED`, `Fuse`. Any other word ahead of four digits is naming something else, and a wrong token here empties the search, so `BatteryHolder_Keystone_1060`, `Crystal_SMD_0603-2Pin`, `Dipole_1090_Arms` and `Jack_1503_02` all stay unread (the last two have the size as a whole second segment, so only the class list refuses them).

Checked against all 15,436 footprints in KiCad 10's installed library: the 152 names shaped `<class>_<4–5 digits>` all carry the Metric companion already, and the new rule changes the output for none of them.

## Verification

- `uvx ruff@0.11.4 check` and `format --check`: clean
- plain lane: 2794 passed, 437 skipped
- native wx controls lane in Docker (Xvfb, CI's exact command): 3199 passed, 3 skipped, 0 failed; the two parameter-colour cases pass again
- new unit coverage: five bare spellings resolve, five look-alikes stay unread, and `expand_footprint("R1", "Resistor_SMD:R_0603")` yields `0603`


z2amiller:
@chmorgan If you could prioritize this one - the regexp collided with some of the new tests (and CI steps) added in the variant, so despite being a clean merge, the CI fails until this goes in.

chmorgan (approved):
Reviewed `d318f85`. The class-restricted fallback restores legacy bare chip sizes without weakening the existing metric-paired matching.

I reproduced both reported native rendering failures against the base helper (one coloured run), then verified both pass on this head with real macOS wx controls. The 80 focused footprint/highlight tests and changed-file Ruff lint/format checks pass; the six positive regression cases fail against the base as expected. Comparing all 15,450 locally installed KiCad footprint names found no output differences.

No blocking findings. Native validation used deterministic matrix fixtures on macOS; Windows/Linux were not rerun locally, and a legacy-board upgrade workflow was not exercised. Current GitHub checks are all successful.

